PR #17, opened 2016-03-09
Enabled committing an empty user_content folder

markwalet:
Small fix that will prevent the empty `user_content` folder from being ignored by Git.
All files in the user_content folder will be ignored.


Attendize:
Thanks for fix @markwalet

PR #19, opened 2016-03-09
Fixed urls for Global Meta file

markwalet:
When you don't run Attendize on a root domain (`http://localhost/attendize/public` for example) the `GlobalMeta.blade.php` file didn't generate the right path to the assets. 
Now using the `asset()` function for the right result.
Note: I also switched the sizes and href properties for the `icon` metatags. Just to have it in the same order as the `apple-touch-icon` metatags.


Attendize:
Nice one,  cheers!

PR #28, opened 2016-03-10
Config update

markwalet:
Also added the ide-helper package from Barry.
Saw that you were using Phpstorm so this package could help you a lot with autocompletion


Attendize:
Nice one! Cheers Mark

PR #31, opened 2016-03-11
Resend Ticket Email

bretto36:
- Added routes
 - Changed Attendees.blade to add a button group instead of just the
message button. Added message button to button group


daveearley:
Nice work! Just tested and it works great. Nice touch with the action dropdown.

PR #34, opened 2016-03-13
Viewing cancelled order

bretto36:
Added Refunded Amount to Order Details view
Removed cancelled tickets from EventTickets View
Corrected first year for CC


daveearley:
Thanks @bretto36 !

PR #37, opened 2016-03-14
Removed duplicate refund code when issuing a partial refund

bretto36:
I also fixed a float comparison bug. When i did a partial refund of 25 (total order 25.99) it worked. Then i tried to do another partial at .99 it kept saying the max i could do was .99 (which is what i was doing). Added rounding to 2 decimal places. If we ever support currencies with more than 2 decimal places this will need to be updated.


daveearley:
Cheers! Nice spot with the floats

PR #40, opened 2016-03-14
Layout fixes

bretto36:
A few layout fixes
Fixed the css for the Check In page so you can actually see the search box. 
Fixed issue when searching for a person already checked it, it wasn't displaying the check mark, but when you clicked it, it would tell you they were already checked in


daveearley:
Cheers @bretto36 !

PR #42, opened 2016-03-16
Updated vendor config files

markwalet:
Do you want to keep the debugbar package?

Edit: Also updated the wkhtml2pdf config file for automatic debugging.


daveearley:
No that's fine to get rid of. Cheers Mark!

PR #44, opened 2016-03-16
View update

markwalet:
Dit some refactoring on a number of views. Did also a partial html5 validation runover and fixed some stuff.


daveearley:
Nice one, cheers!

PR #45, opened 2016-03-17
Add model factories

jncampbell:
- Adds the fzaninotto/faker library
- Adds several model factories to be used during tests.

Please scrutinize the data being generated. Many of the models aren't filled in yet, so I had to look to the migrations file for an idea of what data should be generated by the factory models. If you find any errors, let me know and I'll correct them.


daveearley:
All looks good, Nice one! I'll try hammer out some more later too.

PR #49, opened 2016-03-21
Migration clean ups to ensure rollbacks work correctly.

bretto36:
- correctly removed foreign keys before removing columns
- turned off foreign key checks in initial migration to allow tables to
be dropped


daveearley:
Cheers @bretto36 !

PR #52, opened 2016-03-22
Added css update to make color picker display

bretto36:
Color picker wasn't displaying outside of the panel div as panels have overflow: hidden; by default


daveearley:
Cheers!

PR #64, opened 2016-03-30
Add Event Check-in with QR Code Scanning

mercuryseries:
Add a Web Platform for Event Check-In via QR Code.


daveearley:
Nice work! I tested it on mobile also and it seems to perform well. Cheers!

mercuryseries:
Thanks. You're welcome!

composer.json
```diff
@@ -25,7 +25,8 @@
         "omnipay/paypal": "*",
         "omnipay/bitpay": "dev-master",
         "omnipay/coinbase": "dev-master",
-        "barryvdh/laravel-ide-helper": "^2.1"
+        "barryvdh/laravel-ide-helper": "^2.1",
```
gamerwalt: I thought this should be in require-dev?

PR #67, opened 2016-04-05
Fix issue when creating blank folder with filename

mercuryseries:
Add dirname function in order to avoid to create a blank folder with the filename.


daveearley:
Cheers @mercuryseries !

PR #79, opened 2016-04-20
Postgres compatibility

ylecuyer:
Those are some changes I had to make to have Attendize run with postgresql.

I did not test all the screens/functionalities so additional changes may be required.

### DIFF
--- a/app/Models/EventStats.php
+++ b/app/Models/EventStats.php
@@ -49,7 +49,7 @@ class EventStats extends \Illuminate\Database\Eloquent\Model
     {
         $stats = $this->firstOrNew([
             'event_id' => $event_id,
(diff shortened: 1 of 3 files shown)


app/Models/EventStats.php
```diff
@@ -93,8 +93,8 @@ public function updateTicketsSoldCount($event_id, $count)
     {
         $stats = $this->firstOrNew([
             'event_id' => $event_id,
-            'date'     => DB::raw('CURDATE()'),
-        ]);
```
daveearley: Thanks for this! the only issue is the missing square bracket here.

ylecuyer: Fixed

daveearley:
Thanks @ylecuyer !

PR #89, opened 2016-05-09
Fix for non-unique modal IDs

JapSeyz:
Hey. I noticed that there were several elements with the if "general", and it bugged the account settings modal. I simply added a suffix to prevent this issue from happening again.

Documentation of the issue: [Video](https://www.youtube.com/watch?v=QbeKxNKyZAE)


daveearley:
Nice one, cheers!

PR #100, opened 2016-05-13
I18n

ylecuyer:
**Base for i18n**

Each organiser account has a locale attribute, current page display accordingly to this locale if connected, in english otherwise.
	
Each attendee has a locale attribute (english by default), when sending mail locale should be changed to match attendee locale (see last commit 7413d59).

TODO: replace all the hard coded text with the corresponding `trans()`

In order to generate the locale files use the following command : `php artisan localization:missing -n -b -c -l "■■■■■■■"`

A `dummy` locale has been added and should be used for development purpose only. In this locale every text translates to ■■■■■■■, therefore, given a page it's easy to see what need to be translated.

In order to activate this locale, you must change the locale attribute in the `users` or `attendees` table


daveearley:
Very nice start, cheers!

PR #103, opened 2016-05-21
Fixed installer

ylecuyer:
Added key generation
Added database type selection
Added dependency and optional dependencies


daveearley:
Cheers @ylecuyer !

PR #110, opened 2016-05-22
First revision of Modularity

JapSeyz:
Uses same coding/commenting style.

@TODO is to create a fork of https://github.com/pingpong-labs/modules and change the stubs. Currently, it is necessary to manually extend the MyModuleController and set the routing namespace, I want to do that automatically by pulling it a forked version of PingPong-labs/modules.

There's still a lot of clean-up to be done, but this should serve as a first iteration.
Since there's no admin-panel where the owner of the service can pick which modules are installed, I've opted to let them be installed via composer.


daveearley:
Nice work and very good start! I created a Modularity branch, so can you make a new PR into that?

JapSeyz:
Sure.

PR #113, opened 2016-05-23
First revision of Modularity

JapSeyz:
Now also on the correct branch.

@TODO is to create a fork of https://github.com/pingpong-labs/modules and change the stubs. Currently, it is necessary to manually extend the MyModuleController and set the routing namespace, I want to do that automatically by pulling it a forked version of PingPong-labs/modules.

There's still a lot of clean-up to be done, but this should serve as a first iteration.
Since there's no admin-panel where the owner of the service can pick which modules are installed, I've opted to let them be installed via composer.


ylecuyer:
I don't get the point of this. Is it like a plugin system?

JapSeyz:
Yeah, that's exactly what it is, allows you to extend the Attendize site with some easy to install modules

PR #115, opened 2016-05-29
Ability to refund individual tickets in an order

JapSeyz:
Any order using Stripe can now have individual tickets refunded directly from the order management modal.


ylecuyer:
Please don't commit code formatting changes, where is the actual change in dc93a17 ?

config/services.php
```diff
@@ -36,5 +36,8 @@
 
     'twitter' => [
         'widget_id' => env('TWITTER_WIDGET_ID'),
+    ],
+    'sparkpost' => [
+        'secret' => env('SPARKPOST_SECRET')
```
ylecuyer: I don't think it is related to the refund code

app/Http/Controllers/EventAttendeesController.php
```diff
@@ -662,7 +662,7 @@ public function postCancelAttendee(Request $request, $event_id, $attendee_id)
                 $message->to($attendee->email, $attendee->full_name)
                         ->from(config('attendize.outgoing_email_noreply'), $attendee->event->organiser->name)
                         ->replyTo($attendee->event->organiser->email, $attendee->event->organiser->name)
-                        ->subject('You\'r ticket has been cancelled');
+                        ->subject('Your ticket has been cancelled');
```
ylecuyer: Squash with previous commit

config/services.php
JapSeyz: That's true, it is a general fix for using SparkPost. The TransportManager expects that to be there.

It's a bug that it is missing IMO, and thus is fixed here.

app/Http/Controllers/EventAttendeesController.php
JapSeyz: I am not extremely experienced with Git, and will look up how to do that in the future.

JapSeyz:
Sorry about the formatting change, I didn't catch that. I auto-format it to PSR-2 when I save via PHPStorm, I must've forgotten to turn it off for Attendize on my laptop.

config/services.php
ylecuyer: Then it should be in an independent commit

app/Http/Controllers/EventAttendeesController.php
ylecuyer: You just have to do git rebase -i master and follow onscreen instructions

config/services.php
JapSeyz: I'll remember that.

daveearley:
Hi @JapSeyz , thanks for this. Could you resolve the conflicts on and then resubmit, some files aren't up to date. Cheers!

JapSeyz:
Of course. I will probably do a bit of reworking, and have it save the ticket price, to make it independent of ticket price changes.

However, this will be in two weeks. I have 7 hours of oral exams in the next ten days, that I have to prepare for.

PR #124, opened 2016-06-14
Allow Refunds of individual tickets

JapSeyz:
Can now refund individual tickets, for their CURRENT PRICE. (It doesn't store ticket-price when they're purchased, so they will be refunded the current value that a ticket is set to be sold for).


daveearley:
Cheers! this looks good. I'll give it a test and merge when I'm at a computer.

JapSeyz:
Great, I didn't test it yet, I just moved all the code and have faith in my
skills... Or just forgot it :p let me know if there's any issues.

PR #125, opened 2016-06-14
Fixes Laravel issue with Cookie Middleware

JapSeyz:
The original original cookie middleware is no longer located in app/http/middleware, so the location has been updated to the Laravel default location.


daveearley:
Cheers @JapSeyz !

PR #128, opened 2016-06-15
Allow Ajax Validation errors in Frontend.js

JapSeyz:
For some odd reason, this was left out of the frontend, but implemented in App.js and Backend.js

### DIFF
--- a/public/assets/javascript/app-public.js
+++ b/public/assets/javascript/app-public.js
@@ -20,6 +20,13 @@ $(function() {
                     },
                     error: function(data, statusText, xhr, $form) {
                         $submitButton = $form.find('input[type=submit]');
+
+                        // Form validation error.
+                        if (422 == data.status) {
+                            processFormErrors($form, $.parseJSON(data.responseText));
+                            return;
+                        }
+
                         toggleSubmitDisabled($submitButton);
                         showMessage('Whoops!, it looks like something went wrong on our servers.\n\
                    Please try again, or contact support if the problem persists.');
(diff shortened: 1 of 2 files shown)


daveearley:
Was this causing any errors on your side? I don't remember why but I think it may have been intentionally left out. 

Also, frontend.js is the concatenated file so it's better to make changes in app-public.js then run **grunt js**.

Cheers @JapSeyz !

JapSeyz:
@daveearley Yes.

If you use frontend.js, RequestValidation will not show actual error messages, just an "oops" message, which I don't believe is intentional?

All it does now is to add the help-block to the input if there's an error.

(And also I've added support for input-groups for it).

I'll do a grunt, and update in a sec.

JapSeyz:
@daveearley For some reason, the Session Driver refuse to work with Requests. (also with the proper Web middleware)

Therefore properly submitting a form and doing a page-reload will not show any validation errors, and they were left out when doing and ajax-form. The user wasn't given any kind of feedback on the error they made.

This should fix that.

daveearley:
I had a look there and the validation errors come back as a 200 response so they're currently dealt with in the success callback. The error callback only gets called when there're a server error.

JapSeyz:
I think we're talking misunderstanding each other.

If I do a "non-ajax" form submit, the validation errors session will not be set (I can not tell you why it won't, as I have not been able to figure it out).
Same with old().

The sessions will simply not persist. (I __am__ using the Web middleware, but it could be an issue with the Modules).

TL;DR: Error Validation Sessions from Requests would not be set, but with this script, and submitting through Ajax, the validations will be shown.

JapSeyz:
If we look at line 56 in public-app.js, the function actually exists there.


I just didn't see it because it's only thrown when you return a 200 OK HTTP code.
I prefer to throw the proper 422 HTTP code when there's validation errors.

In the future, both of the calls should be linked to the same method. I'll look at that later.

PR #129, opened 2016-06-16
Started Adding Tests. 

bretto36:
Hi Dave/Others. Just adding this in to get some feedback. 

I've added a couple of tests in and want to know if i'm on the right track. Added an ajax test for creating event. Could you take a look and tell me if it's the way you want to go forward with it, and give me and pointers. 

I've also found a couple of routes that are no longer used (showEditOrganiser, postEditOrganiser), seems to use Customise now instead. 

Found a few issues with Response:: being called which caused errors for me, so switched to response()

Apart from those it's just test changes. I will try and do a couple a day to give us at least some test coverage

### DIFF
--- a/app/Http/Controllers/EventAttendeesController.php
+++ b/app/Http/Controllers/EventAttendeesController.php
@@ -20,7 +20,6 @@ use Excel;
 use Mail;
 use Omnipay\Omnipay;
 use PDF;
-use Response;
 use Validator;
 use Config;
 use Log;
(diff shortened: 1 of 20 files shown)


daveearley:
HI @bretto36 , thanks for this! Looks fine and definitely on the right track. 

I keep meaning to get on top of writing tests but something else always grabs my attention. Iit would be great to get a hand. Cheers!

bretto36:
Testing always gets put by the way side, so valuable but also new features are equally as valuable. Hopefully i can get the main features tested and then we can all add the extra level of tests later

PR #132, opened 2016-06-16
Sticky footer

JapSeyz:
Makes the footer sticky.

(Especially useful for the upcoming modules, as they might not necessarily be full height)
(Also fixes background darkening issue when the content wasn't full height)


daveearley:
Thanks @JapSeyz !

PR #135, opened 2016-06-16
Fixed Edit User

bretto36:
Have fixed the edit user validations and added a test


daveearley:
Nice one @bretto36 , cheers!

PR #138, opened 2016-06-17
Fix sending invites

ylecuyer:


kasbah:
In what cases was is sending invites broken? I have not noticed any problems...

daveearley:
@kasbah In some cases the PDF wouldn't have existed when the email was being sent. 

@ylecuyer thank you!

PR #139, opened 2016-06-19
Replaced manual function with existing one.

JapSeyz:
Function already existed, call that instead of manually writing it out.

### DIFF
--- a/public/assets/javascript/app-public.js
+++ b/public/assets/javascript/app-public.js
@@ -53,24 +53,9 @@ $(function() {
 
                             case 'error':
                                 if (data.messages) {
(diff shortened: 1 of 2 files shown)


daveearley:
@JapSeyz the processForms function here differs from the backend one. As far as I can see this would cause errors to not show on inputs with array names such a **"ticket_holder_first_name.0.21"**

JapSeyz:
@daveearley This should be fixed now. I missed it when I initially ported over the function.

It supports Arrays & Input-groups now.

daveearley:
Cheers @JapSeyz !

PR #142, opened 2016-06-20
Make 1 the default ticket quantity if there is only one ticket type

kasbah:
This seems like a more sensible default. Not sure what the zero option was for though.


kasbah:
I just realized why 0 is an option: when multiple types of tickets are available you would want some to be at "0". I still think having one set to "1" would make sense, especially if there is only one ticket type.

kasbah:
I have changed it to be a default of 1 ticket if there is only one ticket type and back to the old behaviour if there are multiple ticket types.

daveearley:
Cheers @kasbah , the '1' default seems sensible when there's only 1 ticket alright.

PR #144, opened 2016-06-20
Added jquery.payment

kasbah:
and trying to use 'formatCardNumber' on cart number input

I put a big fat [`console.log('HI THERE!')` in the code](https://github.com/Attendize/Attendize/compare/master...kasbah:format-cc-input?expand=1#diff-58a1b04120aebd6d80985875658b5629R98) but it's not getting called. If I do `$('.card-number').payment('formatCardNumber')` on the console then I do get the desired result but I am not sure why the other stuff isn't getting called.


kasbah:
Nevermind, figured it out. 

We could also consider making the expiry input more like the [Stripe demo](http://stripe.github.io/jquery.payment/example/).

daveearley:
Thanks @kasbah !

PR #184, opened 2016-08-29
Fix no tickets selected returns an exception

JapSeyz:
Without this, you woud get an exception further down in the code that would result in an exception when giving a non-set variable to the foreach loop.


daveearley:
Cheers @JapSeyz !

PR #199, opened 2016-09-11
Add user importable calendar to email

jmowatt:
Add the ability to download an ICS file for the event, this link is provided in the ticket confirmation email.


daveearley:
Nice one @jmowatt , thanks for this!

PR #211, opened 2016-10-07
Fix on MIGS payment gateway

jacobcat:
MIGS won't accept order transaction numbers and order descriptions that are not unique, so I made some changes to make sure that the unique transaction number is stored in the order description as well. With this change, users can now successfully make a payment using MIGS.


daveearley:
Thanks @jacobcat !

PR #222, opened 2016-10-13
Update Order.php

jharris70:
Fixed Line 98.


daveearley:
Cheers @jharris70 !

jharris70:
I've got to say the web interface made it easier than I thought it would be to make a quick fix like this. I wonder if I should still learn how to use CLI git properly.

PR #226, opened 2016-10-23
Two bugfixes

nbdd0121:
Display as raw HTML after escaping and nl2br:
With current code, `<br/>` will be displayed instead of a line break. This commit fixes the issue.

Make tickets be sorted in ASC order:
In EventTicketsController::postUpdateTicketsOrder, sort_order is generated in ascending order, however, when displaying tickets, tickets is sorted in descending order. 964e456 tries to solve the issue but it is the other way around.


daveearley:
Cheers @nbdd0121 !

PR #229, opened 2016-10-26
Correctly disable and un-disable fields

nbdd0121:
when toggle between offline and online payment methods.


daveearley:
Cheers @nbdd0121 !

PR #243, opened 2016-11-09
Update EventCheckoutController.php : stripe receipts.

jharris70:
Added 'receipt_email' to the Stripe charge creation request so customers get a payment receipt as well as their tickets. Also removed a blank line.

I would still like to see the $order->order_reference in the description, but I'm not sure if $order has that property at this point: line 330 of EventCheckoutController.php.


daveearley:
Thanks for this @jharris70!

PR #253, opened 2016-11-15
Removed non-unique email templates

JapSeyz:
There were three of the same email templates, this has been changed to just one.


daveearley:
Cheers @JapSeyz!

PR #259, opened 2016-11-19
Uniform Organiser Image Upload

JapSeyz:
* Preserve Aspect Ratio
* Use the same size no matter where the upload is happening


daveearley:
Nice one! cheers @JapSeyz !

PR #260, opened 2016-11-20
Update InstallerController

MsFoued:
create database automatically


daveearley:
Hi @MsFoued ,

Thanks for the contribution! The reason we don't create a DB during the install is because the install could possibly fail if the user doesn't have permissions to create a DB. 

I'll happily merge the artisan command PR you created however.

Many thanks!

PR #262, opened 2016-11-20
Update Kernel.php

MsFoued:


JapSeyz:
Add this and #261 into a single PR, probably with #260 

It makes no sense to have three different PR's for features that depend on each other.

You can add commits to a PR after it has been made.

PR #265, opened 2016-11-24
Allow the creation of blank printable tickets

JapSeyz:
Also append all imported users to the same order, for easier deletion if the import was a fluke.

* Removed the 0.00/free and replaced it with the real charge on tickets sent to imported users


daveearley:
Conflict here @JapSeyz :-)

JapSeyz:
Whoops, I'll take a look

JapSeyz:
@daveearley There you go :)

PR #267, opened 2016-11-28
Reflect Attendee cancellation in event stats

JapSeyz:
When an Attendee is cancelled, also remove the Attendee from the stats, to prevent errornous amounts for the event being shown.


daveearley:
Cheers @JapSeyz !

Issue #283, opened 2017-01-01
php 7.1/ mcrypt_encrypt()/ cipher problem


KlausGPaul:
I was struggling recently to get attendize up and running in docker again after it had been working a while ago. It seems the issue is related with an upgrade from php 7.0 to 7.1, see [here](http://stackoverflow.com/questions/41031076/php7-1-and-laravel-5-3-function-mcrypt-get-iv-size-is-deprecated) for a similar error.

While I was not changing the cipher in .env (which would be one thing) I was changing Dockerfile-php from
FROM php:fpm
to
FROM php:7.0-fpm

After that, attendize was happily working again.

RandomHeroo:
You are right im having the same issue in Fedora 25.

` ErrorException in McryptEncrypter.php line 46: Function mcrypt_get_iv_size() is deprecated

    in McryptEncrypter.php line 46
    at HandleExceptions->handleError('8192', 'Function mcrypt_get_iv_size() is deprecated', '/var/www/html/attendize/vendor/laravel/framework/src/Illuminate/Encryption/McryptEncrypter.php', '46', array('key' => '$�j�8Yy�����YXz�ԧ�ȧ����?�', 'cipher' => 'rijndael-128')) in McryptEncrypter.php line 46
    at McryptEncrypter->__construct('$�j�8Yy�����YXz�ԧ�ȧ����?�', 'rijndael-128') in compiled.php line 7789
    at EncryptionServiceProvider->getEncrypterForKeyAndCipher('$�j�8Yy�����YXz�ԧ�ȧ����?�', 'rijndael-128') in compiled.php line 7781
    at EncryptionServiceProvider->Illuminate\Encryption\{closure}(object(Application), array()) in compiled.php line 1353
    at Container->build(object(Closure), array()) in compiled.php line 1309
    at Container->make('encrypter', array()) in compiled.php line 1871
    at Application->make('encrypter') in compiled.php line 1410
    at Container->resolveClass(object(ReflectionParameter)) in compiled.php line 1387
    at Container->getDependencies(array(object(ReflectionParameter)), array()) in compiled.php line 1373
    at Container->build('Illuminate\Cookie\Middleware\EncryptCookies', array()) in compiled.php line 1309
    at Container->make('Illuminate\Cookie\Middleware\EncryptCookies', array()) in compiled.php line 1871
    at Application->make('Illuminate\Cookie\Middleware\EncryptCookies') in compiled.php line 9975
[...]

PR #294, opened 2017-01-20
#283 Cipher can be set in ENV file

Tastenbimbo:
#283 
* Set AES-256-CBC as default cipher in ENV file
* config/app.php edited 'cipher' => env('APP_CIPHER', 'MCRYPT_RIJNDAEL_128'),


daveearley:
Many thanks @Tastenbimbo!

PR #310, opened 2017-02-06
fix docker-compose setup configurations

tristanguo:
- Change .env.example to use postgres, cause in docker compose file, postgres sql is being used
- Change php base image to php:7.0-fpm, cause using latest will have compatible issue
- Rewrite docker-compose file to version 2
- Modify readme docker section, cause some of the steps are wrong


KlausGPaul:
Thanks from me, too!

PR #343, opened 2017-04-26
Sub Tickets Option on Tickets

vengatowen:
Hi, I'm not sure is this is the right place to request feature help. 

I would like to create a sub ticket with selection options (checkbox./option) with the main ticket. How to add?


sagargulati:
No support & updates?

PR #348, opened 2018-02-16
QR-code check-in

szpiku:
Hello,

I've got a problem with QR corde check in on smartphone.
When I going to check in and turn on scanner, there is only the background with frame and 'scanning...' but the camera wont turnon.
I've always turn on permision to access the camera. 
Check on few different webclients (like ff/chrome/build in)

But still it works on the PC.


codehakase:
@saifur1408 edit the `config/database.php` file with your database credentials (set this up first), then re-run `php artisan migrate`

PR #375, opened 2018-05-21
Set users/customers as not checked-in by default

theLesa:
Hello,
I would have a suggestion. It would probably be better to set the users/customer as "checked-out" by default at the beginning. As I could see in the demo, all new customers will automatically be checked in, which is not what you want as an organiser.

Thank you for great work


samdb:
Incorrect pull request

PR #392, opened 2018-07-04
Fix docker install process

jeremyquinton:
Updated the default Docker configuration so that an Attendize development environment can be built straight out the box.


jeremyquinton:
@samdb when you say the autofill for Docker do you mean the pre populated database details on the installer page. 

It would be possible to detect if the installer is running on Docker and then fill out the details. 

I haven't seen that error before but going to do a fresh run on my wifes mac to see if I can replicate it.

Issue #400, opened 2018-07-12
Official Docker Image


Leen15:
Hi to all!
Thanks for the great work you did with Attendize, it looks like a great product.
I would like to use Attendize in docker, but to be honest the docker-compose structure is a pain.
We would like to deploy this in a cluster (Rancher and/or kubernetes) but this is not possible if I have to build manually the repo.
Usually, every docker implementation should have an official image, something that can be started and runs out of the box.
I tried to make this working, but there are lots of things wrong with the current implementation.
First of all, the .env file shouldn't exists. In every orchestrated system ENVS are passed on container start and shouldn't exist inside the repo during the build process.
I developed in the past a laravel project with docker support, and my approach (that it's not perfect of course) was to generate the env file from a base file (see https://github.com/Leen15/rancher-wrapper/blob/master/start.sh#L4-L5 ). This allow to declare outside the container what needed and start it with no issues from a builded image.
Second thing, the composer part should stay inside the dockerfile, so no need to manually do it when container start. Here you could consider the docker cache system, so copying only composer files before call the composer install can help to not do a full install every time the image builds (but only if something changes in composer dependencies). 
This is an example:
![image](https://user-images.githubusercontent.com/3629071/42649942-5cd6934a-860b-11e8-9bce-aaa9bec21f01.png)

Last of all, I cannot understand why you separate nginx from php in two different containers, is this really required? could an apache container with php or a base image with nginx+php be more easy to use and maintain? This could also help to scale easier when traffic grows.

I know this project is in an alpha stage, but an official image can be very useful for lots of people that would like to try this easily and help with some testing and bug reports.

Leen15:
Hi @jeremyquinton thanks for your answer. 
I made a PR about my current implementation so maybe this can be useful as a starting point for a better implementation on the offical repo.

jeremyquinton:
@Leen15 I have had a look at your PR and I think it has some definite improvements that we can adapt into the official repo for Docker setup. I will get back to you when these are merged into master.

Leen15:
@jeremyquinton  In my tests I saw that the image has some issues if it stays behind a proxy with SSL termination. All assets links are generated in http so they don't load due to mixed content policy.
I added a new env "FORCE_HTTPS" for generate all assets with https.
I don't know if it's the right way (I saw that change a lot based on laravel version) but it's a couple of hours that I'm trying and it works.
[...]

PR #401, opened 2018-07-13
docker improvements for production enviroment

Leen15:
I made some improvements on docker installation for a "production" environment (#400).
Now envs are loaded outside the container, no need to manually create a .env file inside the repo.
I created a .env.dist file, this is used for references on what envs are copied inside the container folder at startup.
Dockerfile is only one, it uses `ubuntu 16.04 + apache + php7.1 + mysql-client`.
`Composer install` now runs during the image build, so no need to manually run it after the container start.
I would like to use the docker cache system for avoid to run at every build but I found lots of references under the /app folder so I cannot copy only composer files (json, lock, package).
I added two bash script:
`generate_env.sh`: this script search in the environment if .env.dist envs exists and prepare a list of valorized envs to use for the .env file
`start.sh`: this script runs at container startup, check for .env file and if not exists it creates a new one on the project path, generate the APP_KEY env and reload the cache config. This script also check for a `WORKER` env, used for launch the queue demon instead of apache. So with this logic the worker container generates the .env file for itself too and starts correctly. I also add the support for a unique APP_KEY env for both web and worker, so if it exists it doesn't generate a different one.

The resulted docker-compose is really simple and you can launch with `docker-compose -f docker-compose.production.yml up`


bzg:
Hi, just a big +1 on this PR. I've installed Attendize and I stumbled on the `mixed content` bug, which prevents me from using Attendize in production.

jeremyquinton:
@Leen15 I can merge this https://github.com/Leen15/Attendize/compare/master...Leen15:feature/add_force_https_env?expand=1 if you open a separate PR for it. 

The problem with the way you have set this up is that you are running multiple applications/processes in a single container. I realise this is just development but I'd prefer to adhere to the patterns and intentions of the tool at hand which is docker.

The extra bits you have added make total sense. Would you be able to adapt them to the current docker setup? Id be happy to do that. If we have less steps for installation thats an improvement.

Leen15:
Hi @jeremyquinton, the implementation I have proposed, perfectly works with a single application for container. To be honest we use it in this way.

The only difference respect to the normal way is that you can pass an ENV to use the container as queue processor instead of webserver.

This implementation doesn't mean that the container does multiple things, it only support to run in multiple ways to avoid creating several different images for the same repo that has at the end similar scopes.

bagage:
Is it planned to merge this any time soon? Deploying Attendize via docker-compose is not an easy task, it would be great to have a production-ready image for it.

Bellusterra:
@Leen15 Thank you for the docker image, I'm using it right now, and works great :) When installing just noticed a missed space on line 112 of the readme.md file which reads "comfortableon" instead of "comfortable on". Not super important just thought I would point it out.

Bellusterra:
@bzg I had the same issue with the mixed content @Leen15 actually put out a fix for this which fixed my issue and forces https you can find it [here](https://github.com/Leen15/Attendize/compare/master...Leen15:feature/add_force_https_env?expand=1).
also I realize it might be self explanatory but when adding the  FORCE_HTTPS= to the .env file make sure to fill it in :) FORCE_HTTPS=https://yoursite.com

bzg:
Hi @Bellusterra, thanks for the PR and for lletting me know, I will test this over the week-end.

@jeremyquinton My team and I worked on the french translation but I cannot really use Attendize until the docker image is usable over `https` - if you can have a look at the PR, that'd be great!

bzg:
Just a quick heads up to know the status of this issue: how close are we to be able to run a docker image for production environments? Thanks @jeremyquinton for any feedback.

bzg:
Another quick heads up @jeremyquinton - is it okay now to run Attendize in production via docker or is this issue still preventing it?

We are a small team at https://www.etalab.gouv.fr and I'd very much like to use a free software to let us organize events. So far we are stuck with closed solutions.

Thanks a lot.

bzg:
Hi @jeremyquinton - can we get a quick update on the status of this PR? Is it obsolete now?

A short reminder about the context: I'm working for the french central administration and I'm looking for an open source solution to manage subscriptions to our events. I would love to be able to test Attendize.

Thanks!

johannac:
@Leen15 Thanks for your contribution! I've rebased your changes onto the current master branch, and will merge them into our develop branch. I'll make a few tweaks and changes to update and finalise some things, and then it will get released in a couple weeks, by around June 1st.

johannac:
For anyone still following this, here is the most recent PR with updated docker config for both nginx and apache https://github.com/Attendize/Attendize/pull/802

PR #424, opened 2018-08-15
Upgrading to Laravel 5.6

jeremyquinton:
1) Changed method call from lists to pluck.
2) Use session method put instead of session method set.
3) Removed references to artisan optimize command.
4) Added additionl check to base controller if auth user not set redirect to login.
5) Flush the session on logout.
6) Updated Event and Route service providers to work with Laravel 5.6.
7) Added new default logging config file.
8) Bump PHP versions from 7.0.30 to 7.1.20 in Docker containers. 
9) Added missing translations.
10) Fixed some issues with the templates.


emergingdzns:
Does this mean that the system is ready to be updated to L5.6? If I run a git pull will it update? Or do you have any specific instructions on the update process?

publicarray:
@jeremyquinton not sure if you figured it out on your own by now, but a `composer install` or `composer update` is required after the `git pull` to pull in L5.6 and update other dependencies.

Issue #306, opened 2017-02-03
Transaction rollback if email not sent after ticket purchase


jshailes:
Upon successfully purchasing a ticket via Paypal, the pdf is generated and attempted to be sent out via email. If the email fails to send due to bad credentials it seems the order data is not persisted to the database. Nginx error log attached.
[emailNotSent.txt](https://github.com/Attendize/Attendize/files/749918/emailNotSent.txt)

daveearley:
Hi @jshailes, The problem seems to [occur here ](https://github.com/Attendize/Attendize/blob/87fb3ad4b2352cdd32eae2ff8dbf1fdcc747565e/app/Http/Controllers/EventCheckoutController.php#L610-L610). Usually the send email/PDF generation is passed off to a queue, so if something fails the order is still generated. Unfortunately if Attendize is set to run without a queue bad email credentials etc. will cause the whole order to fail. In v2 I'll revisit this to make sure orders won't be lost :+1:

jeremyquinton:
Once #429 is merged what will happen is the order will be created in the database. However if the ticket can't be generated and the attendee email can't be sent this does not affect the order being persisted in the database. Going to close this ticket for now.

jeremyquinton:
Closing this issue as #429 has been merged.

Issue #275, opened 2016-12-09
Booking not registering


angrydogmedi:
Getting error message
"Whoops!, it looks like the server returned an error. Please try again, or contact the webmaster if the problem persists." after submitting a booking this is both with free and paid events,

With paid events(using stripe) the payment does go through, just nothing registering on the system,

it has me at a loss why, no logs are registering either and debug is set to true, 
im at a total loss

jharris70:
I have a similar issue: payments would be processed, and the Whoops error appeared. I figured it was something that I changed when I wasn't looking! The absence of the error log is disappointing, though.

I added an alert panel to the payment page telling people of the false positive error and to wait for a payment receipt before trying again, but some people still just don't read and we ended up refunding multiple charges.

ungelovende:
I'm having the same problem. And when I check the Chrome console I get a Error 500 Internal server error when I try to buy a ticket, the same error even if I try with a free ticket.
But I do not understand why it's a error 500 error, everything els working?

It happen when I use Stripe payment or if I use Offline payment.

daveearley:
Hey @ungelovende ,

Try setting **APP_DEBUG=true** in your **.env**, you should see the full error then.

ungelovende:
@daveearley I've already set true - but doesn't show any logs in "storage/logs"

daveearley:
Is there any error other than the one above coming back from the ajax call in the dev tools?

ungelovende:
No. Just som warnings about some decoding fonts.

I get the same error in attendize when I try to invite a person with a ticket.
dev console show: "POST http://domian.tld/event/12/attendees/resend_ticket 500 (Internal Server Error)"

daveearley:
In that case it could be that your email setup is the problem as there are email being sent when you carry out those actions. Can you post the full response from the AJAX calls?

ungelovende:
When I try to resend a ticket for invitee it says "Whoops!, it looks like the server returned an error." in the browser, the dev console in chrome say "jquery.js:8630 POST http://domain.tld/public/event/12/attendees/resend_ticket 500 (Internal Server Error)"
dev console:
POST http://domain.tld.no/event/12/attendees/resend_ticket 500 (Internal Server Error)
send @ jquery.js:8630
ajax @ jquery.js:8166
$.fn.ajaxSubmit @ backend.js:4179
doAjaxSubmit @ backend.js:4776
dispatch @ jquery.js:4435
r.handle @ jquery.js:4121

[...]

PR #429, opened 2018-08-21
Handle errors independently of one another 

jeremyquinton:


jeremyquinton:
In the previous codebase it would attempt to

1) add the order to the database
2) fire off the events to create tickets and send emails.

There was a try catch block around the functionality in point 1 and 2.

However on point 2 if the job to create the ticket failed (permissions issue etc) and an exception was thrown it would rollback the order.

Exceptions thrown in point 1 and 2 above should be handled independently of one another. Additionally the items in point 2 are intended to run 

Improving the logic e.g. moving the event firing to after the order has been placed removes weird race conditions and errors that existed in Attendize. This solves #306 and #275 amongst other things.

PR #446, opened 2018-09-02
445 fix surveys bug after updating to L5.6

heilgar:
HI, I found one bug after I updated to larval 5.6.
Files which related to surveys were not updated from ->list to ->pluck.


heilgar:
@jeremyquinton please have a look

jeremyquinton:
@heilgar thanks for the fix I have merged this

Issue #428, opened 2018-08-20
Customize Event Start/End format hard coded


emergingdzns:
It appears that the datetime format in the Customize Event General tab for start/end dates and time is using a hard coded "d-m-Y H:i" format. I checked the MyBaseModel and I can see that a format may be passed into the function, but the EditEventForm.blade.php partial isn't passing in a format. Being in the US, we tend to use "m-d-Y h:i a". I added this to the blade as the second parameter in the getFormattedDate function. I then attempted to save and got a server error related to Carbon saying it couldn't parse the date. I checked the EventController at line 230 and it shows this:
```
$event->start_date = $request->get('start_date') ? Carbon::createFromFormat('d-m-Y H:i',
            $request->get('start_date')) : null;
```
Again the date format is hard coded and not set to use the configured date time format.

So, in the blade, I added `config('attendize.default_datetime_format')` to the getFormattedDate calls. I then changed the `d-m-Y H:i` references in the EventController to `config('attendize.default_datetime_format')` and it works great.

Also, in order to correct the creation of new events to follow the same function, in the MyBaseController, I changed these lines:
```
				'DateFormat'          => 'dd-MM-yyyy',
				'DateTimeFormat'      => 'dd-MM-yyyy hh:mm',
```
to
```
				'DateFormat'          => config('attendize.default_date_format'), //'dd-MM-yyyy',
				'DateTimeFormat'      => config('attendize.default_date_picker_format'), //'dd-MM-yyyy hh:mm',
```

Hopefully this helps others with the same problem I had.

jeremyquinton:
@emergingdzns thanks for bringing this up. I'll try get this added as an enhancement at some point.

emergingdzns:
Oh one more area that needed to be modified was the event view page where it shows the date and time at the top. It was hard coded for a different date format than what we wanted.
The file is resources/views/Public/ViewEvent/Partials/EventHeaderSection.blade.php

Since the display formatting for dates and time is different than what we want in text entry fields, I created three new config settings in the attendize.php file:
```
	'display_datetime_format' 	=> 'M j, Y h:i a',
	'display_date_format'		=> 'M j, Y',
	'display_time_format'		=> 'h:i a',
```
I then made the following changes to the view file mentioned above:
Line 27: `{{ $event->start_date->format(config('attendize.display_datetime_format')) }}`
Line 32: `{{ $event->end_date->format(config('attendize.display_time_format')) }}`
[...]

PR #447, opened 2018-09-02
Date/Setting modifications and BluePay gateway

emergingdzns:
In this PR I've included several items:

1. I've set a few of the attendize.php config items to be controlled by the .env file. The reason being is that I think it would be nice if when users go to update attendize that their customizations to these settings not get wiped out. By putting them in the .env we can assure that this won't happen. There are probably more that we should do this with, but these were the changes I made to my file that I don't want to have to redo if I do a merge from the main Attendize repo.

2. BluePay is an available payment gateway. I do NOT recommend using BluePay because it is not a tokenized payment system, but it's what my client is using. In order to make this work I had to make several changes:
    a. In the views, I added an if around the js script tags to make sure that the payment gateway is Stripe. When the stripe js is called without the keys it errors and buggers up the rest of the code.
    b. Also in the views, since BluePay is not a tokenized JS form like Stripe, I needed to remove the "ajax" class from the form and add "name=" to the form elements so that the data submits. There's probably a better way to do this but this is what came to mind.
    c. In the checkout process, when it called `$this->completeOrder($event_id)` it's assuming that the call is an AJAX call. But with BluePay it's not. So I added the same check as in the views to make sure the payment gateway is stripe and if not then it adds `false` to ensure that the page does a redirect, which was already programmed into the code.

3. Added `use Log;` to the SendOrderNotification file as it was throwing errors because of the Log::info calls in the code.

4. I fixed the references to the date formatting throughout the views and the JS to use the config formats as outlined in issue #428 . Ideally these should actually be helper calls rather than config since we don't necessarily want to use the default settings in case a user has a customized date/time format. We really should have some sort of configuration screen for the account owners to manipulate their settings for date_format_id and datetime_format_id. If it's there somewhere I missed it. But we should also allow the user to add new formats and select them. This way a helper call would fetch the settings of the account owner and use their preferred settings rather than the defaults. But this is a step in the right direction getting it to not use a hard coded date format that could get overwritten with a pull/merge.

Feel free to use whatever you want from this.


emergingdzns:
I have now added more to this PR. Included in the latest commits is the ability to manage users and isolate them to specific organisers. This involved the following changes:

1. I added a new js library called bootstrap-multiselect. I used bower to install it but I had to manually copy the code into the backend.js file and had to add the css to the compiled css file in assets. If there's a better way to do this, I'm open to suggestions. I assume you are using some type of compiler but I couldn't figure it out.

2. I shifted a few things around so that the add user row is inside of `tfoot` instead of `tbody` so that now when you add a user it will place it at the end of the list. In the current build it just throws a message at the top of the screen.

3. I've added buttons for editing and deleting users. To edit, I had to change some of the core functionality to look to see if a user id is in the url. If it is, then it checks to make sure the user has the authority to edit the user (must be an admin). To delete, it also verifies authority and that the user isn't trying to delete themselves and then removes the row after deleting the user.

4. I added the section for managing the organisers they are assigned to the edit modal for the users.

5. I added the deleteUser route.

6. Created a migration for the pivot table for organiser_user connection.

7. Modified all views to block non-admins from viewing areas they should not be allowed to view. For instance, if a user is assigned to an organiser, they are able to view the data around events. They can not customize the events or tickets, nor can they create tickets or events. They are allowed to do checkin, cancel/refund tickets, view stats etc. If a user is NOT assigned to an organiser they are considered an "Admin" and able to do anything in the system that users can do in the current version. This way, when adding this change, it won't break existing user's functionality. Also, when users are assigned to organisers it will only allow them to view their organisers, so the drop down in the top right changed to only offer their assigned organiser(s). Same is true of the post login screen where the user must select the organiser to view.

8. Added a couple new methods to the User model: `isAdmin()` and `isOrganiserUser($organiserId)`. The first one checks to see if they are and Admin by not being assigned to any organiser or that they are `is_parent`. The latter method simply runs through the users organisers list and checks to see if the organiser id passed in matches. If so, returns true.

Please let me know if you have any questions or would like me to make any changes.

publicarray:
There are some good changes here, do you think it is possible to create a separate PR for each major feature? this would increase the chance that it will be merged quicker. Unfortunately there was a force push on the master branch so this PR needs to be rebased. For an example see https://github.com/Attendize/Attendize/pull/483#issuecomment-436064180

PR #448, opened 2018-09-03
Fix issue where taxamt not set after import

scottybo:
Fixes server error when trying to import a CSV of invitees


scottybo:
Added a fix for the deprecated "lists" function (checked the rest of the repo and this seems to be the only instance of lists())

Issue #327, opened 2017-03-05
 php artisan route:list error


Tyga76:
when I run php artisan route:list i get the following error. it seems the ApiBaseController has an issue when you run from command line..Any ideas?

[2017-03-05 08:52:00] production.ERROR: exception 'ErrorException' with message 'Trying to get property of non-object' in /Users/tayganpillay/Dev/officialevents/stubs/app/Http/Controllers/API/ApiBaseController.php:15
Stack trace:
#0 /Users/tayganpillay/Dev/officialevents/stubs/app/Http/Controllers/API/ApiBaseController.php(15): Illuminate\Foundation\Bootstrap\HandleExceptions->handleError(8, 'Trying to get p...', '/Users/tayganpi...', 15, Array)
#1 [internal function]: app\Http\Controllers\API\ApiBaseController->__construct()
#2 /Users/tayganpillay/Dev/officialevents/stubs/vendor/laravel/framework/src/Illuminate/Container/Container.php(779): ReflectionClass->newInstanceArgs(Array)
#3 /Users/tayganpillay/Dev/officialevents/stubs/vendor/laravel/framework/src/Illuminate/Container/Container.php(629): Illuminate\Container\Container->build('App\\Http\\Contro...', Array)
#4 /Users/tayganpillay/Dev/officialevents/stubs/vendor/laravel/framework/src/Illuminate/Foundation/Application.php(697): Illuminate\Container\Container->make('App\\Http\\Contro...', Array)
#5 /Users/tayganpillay/Dev/officialevents/stubs/vendor/laravel/framework/src/Illuminate/Foundation/Console/RouteListCommand.php(165): Illuminate\Foundation\Application->make('App\\Http\\Contro...')
#6 /Users/tayganpillay/Dev/officialevents/stubs/vendor/laravel/framework/src/Illuminate/Foundation/Console/RouteListCommand.php(146): Illuminate\Foundation\Console\RouteListCommand->getControllerMiddleware('App\\Http\\Contro...')
#7 /Users/tayganpillay/Dev/officialevents/stubs/vendor/laravel/framework/src/Illuminate/Foundation/Console/RouteListCommand.php(118): Illuminate\Foundation\Console\RouteListCommand->getMiddleware(Object(Illuminate\Routing\Route))
[...]

PR #471, opened 2018-09-22
Improve upstream

publicarray:
I've squashed some bugs, changed min pass length to 8 chars, removed old code ~~and have implemented the DEFAULT_DATETIME_FORMAT option and introduced a new DEFAULT_DATEPICKER_FORMAT option.~~ this is now in #483. These settings aren't perfect but I think it it a big improvement to what is was before.

Sorry that this is a bit much but I've already squashed things from my master branch to ease the code review.

Happy to receive feedback 😄


resources/views/Shared/Layouts/Master.blade.php
```diff
@@ -130,7 +130,7 @@ class="loadModal editUserModal" href="javascript:void(0);"><span class="icon ico
     $(function () {
         $.ajaxSetup({
             headers: {
-                'X-CSRF-Token': "<?php echo csrf_token() ?>"
+                'X-CSRF-TOKEN': $('meta[name="_token"]').attr('content')
```
jeremyquinton: @publicarray can you explain what this change does?

app/Http/routes.php
```diff
@@ -384,18 +384,6 @@ function ($event_id) {
...
-             */
-            Route::get('{event_id}/tickets/questions', [
-                'as'   => 'showTicketQuestions',
-                'uses' => 'EventTicketQuestionsController@showQuestions',
-            ]);
-            Route::post('{event_id}/tickets/questions/create', [
-                'as'   => 'postCreateQuestion',
-                'uses' => 'EventTicketQuestionsController@postCreateQuestion',
-            ]);
-
```
jeremyquinton: @publicarray why did you remove these routes

resources/views/Shared/Layouts/Master.blade.php
publicarray: This is more of a personal thing but I was initially going to convert the php to blade `{{ }}` for constancy.  Later I looked at the [Laravel Docs](https://laravel.com/docs/5.7/csrf#csrf-x-csrf-token) and this is how they have it. I image it might be better because we remain in the JS context and don't switch to php/blade. If you like I can remove this change. I don't have a strong option on this one but I thought going with the convention might be better. (I image it is the convention when people copy from the documentation 😏)

app/Http/routes.php
publicarray: Unfortunately [EventTicketQuestionsController doesn't exist](https://github.com/Attendize/Attendize/search?q=EventTicketQuestionsController&unscoped_q=EventTicketQuestionsController) yet and it fixes https://github.com/Attendize/Attendize/issues/327.

jeremyquinton:
@publicarray I went through this again I can't accept all these changes. I'm going to have to create a new branch and cherry pick the best parts out. 

Are you happy to update this PR removing certain things I mark as remove?

publicarray:
Yea sure I need to fix up the merge conflict (as expected) anyway.

publicarray:
I think I'll mode the date time stuff to a separate PR

PR #475, opened 2018-09-28
fix bug where postMessageOrder gets datas of orders table

dsnap1993:
The function postMessageOrder should get datas from Orders not attendees table but orders table on line 437.
So I have fixed it.

I would be happy if I could get your feedback.


jeremyquinton:
@dsnap1993 thats correct. Weird how this was wrong will merge your PR thanks very much

PR #476, opened 2018-09-29
Fix xss vulnerability

jeremyquinton:
Thanks @publicarray for reporting. 

This PR fixes a possible xss attack where an attendee could inject html via event ordering. 

1) Remove as much html from translations so that we can use {{ }} instead of {!! !!}
2) strip_tags on user input. 
3) escape all user input properly.


publicarray:
Hi Jeremy,

This looks stellar. Thanks for working on this so swiftly.

Unfortunately there is still one in the QR scanner see: https://github.com/publicarray/Attendize/commit/a3eaf406b5741f91675066bdba08f24f8d9b422e and https://github.com/publicarray/Attendize/commit/cb95dc39c5db49a0c262f1f2a789e661f5d5d331
(I've only just pushed it to github now since this PR is public). The name I used to test is `<a onmouseover="alert(document.cookie)">xxs link</a>` which requires a mouse hover over the link.

Since these changes also remove the bold styling you could do `@lang("ManageEvent.edit_attendee_title") <b>{{ $attendee->full_name }}</b>` instead of `{{ @trans("ManageEvent.edit_attendee_title", ["attendee"=> $attendee->full_name]) }}` if you wish.

As for the organiser about section ([OrganiserHeaderSection.blade.php](https://github.com/Attendize/Attendize/blob/master/resources/views/Public/ViewOrganiser/Partials/OrganiserHeaderSection.blade.php) `{!! $organiser->about !!}
`) I leave it to you to decide whether the content is trusted or not.

publicarray:
I recommend you merge this and #478 very soon since this has been public for nearly 3 days. It would not surprise me if someone is crawling GitHub to look for ‘xss’ in PRs or commits. Once it’s merged we should get people to upgrade to the newest version of attendize as fast as possible.

publicarray:
That was quick 👍 Nice!

PR #478, opened 2018-10-02
Fix vulnerabilities

publicarray:
This PR is a continuation from #476

### DIFF
--- a/app/Http/Controllers/EventCheckInController.php
+++ b/app/Http/Controllers/EventCheckInController.php
@@ -185,7 +185,9 @@ class EventCheckInController extends MyBaseController
 
         return response()->json([
             'status'  => 'success',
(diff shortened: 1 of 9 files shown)


jeremyquinton:
@publicarray I was trying to test these changes this morning and the scanner just hangs on iphone 6 chrome safari with a blue screen. What device are you testing on?

publicarray:
I'm on a MacBookPro and I used Google Chrome to test. I've noticed on the laptop sometimes a full system restart is necessary for the isight camera to work. I think there is a bug when my laptop goes to sleep and it stops the isight to start up again. (photo booth dosn't work either)

publicarray:
Did the iPhone 6 work before, without these changes?

jeremyquinton:
@publicarray I'm not sure let me do some testing and then we can get this merge today

PR #515, opened 2018-11-07
Spanish translation

CristianDeluxe:
Hello, I have done the Spanish translation, I have translated only the strings that are available, until more of them can be translated.

Greetings


jeremyquinton:
@CristianDeluxe thanks for the PR.

PR #522, opened 2018-11-18
License URL should be license.html not licence.html (404 page not found error)

bagage:


renepardon:
https://www.dict.cc/?s=licence

https://www.dict.cc/?s=license

Both are valid.

bagage:
Yes but not the link, one goes to 404 page.

PR #534, opened 2018-12-07
Prevent duplicate queries

AdrienPoupa:
The following commit prevents two duplicate queries:

`select count(*) as aggregate from organisers where ((organisers.account_id = 1))`

runs twice because we have `Organiser::scope()->count()` twice in `FirstRunMiddleware`

`select sum(quantity_reserved) as aggregate from reserved_tickets where ticket_id = 1 and expires > '2018-12-07 00:40:19'`

runs 2 or 3 times in event/{id}/tickets because the function `getQuantityReservedAttribute` is called several times and call the database each time the attribute is requested. I suggest that we create an attribute directly in the Ticket model, set it to its actual value once and then serve it directly (soft caching).


johannac:
Due to merge conflict I had to merge this manually on the command line, but it's now in the develop branch.

PR #549, opened 2019-01-09
Allow offline payment when no gateway is defined

DODMax:
This PR does the following:
- Fix checkout page cannot be displayed if no payment gateway is defined
- Force offline payment when no payment gateway is defined (offline payment needs to be enabled)
- Proper error when no payment gateway is defined and offline payment is not enabled
- Fixed a couple of code typos

I tried to modify as few lines of code as possible to keep the same workflow

Thanks and cheers!


jeremyquinton:
@DODMax thanks for this fix. Works nicely essentially allowing an event to accept only offline payments if they want to.

Issue #524, opened 2018-11-20
Installation coming failed


RameshParihar:
Getting following error on step 7 of the installation 
"production.ERROR: Use of undefined constant STDIN - assumed 'STDIN' (this will throw an Error in a future version of PHP) {"exception":"[object] (ErrorException(code: 0): Use of undefined constant STDIN - assumed 'STDIN' (this will throw an Error in a future version of PHP) at C:\\xampp\\htdocs\\event\\vendor\\symfony\\console\\Helper\\QuestionHelper.php:111)"

Kindly provide help on this!!!! Attaching the log file as well. 

Things tried so far:
Downgrading from 7.2 to 7.1.3 still facing the same issue
giving right to all of the folder
fresh installation

RameshParihar:
[laravel.log](https://github.com/Attendize/Attendize/files/2600359/laravel.log)

Issue #458, opened 2018-09-09
Installation error


blagoeres:
I followed the steps in the installation docs and I get this error when pressing "INSTALL ATTENDIZE" button.

My software stack is:

- CentOS 7.5
- Nginx version 1.12.2
- MariaDB version 15.1
- PHP version 7.2.9 form Webtatic YUM repo.
- Composer version 1.7.2
- Attendize version 1.1.1

Here is the `laravel.log` file.

```
[2018-09-09 12:13:22] production.ERROR: Use of undefined constant STDIN - assumed 'STDIN' (this will throw an Error in a future version of PHP) {"exception":"[object] (ErrorException(code: 0): Use of undefined constant STDIN - assumed 'STDIN' (this will throw an Error in a future version of PHP) at /var/www/attendize/vendor/symfony/console/Helper/QuestionHelper.php:100)
[stacktrace]
#0 /var/www/attendize/vendor/symfony/console/Helper/QuestionHelper.php(100): Illuminate\\Foundation\\Bootstrap\\HandleExceptions->handleError(2, 'Use of undefine...', '/var/www/attend...', 100, Array)
#1 /var/www/attendize/vendor/symfony/console/Helper/QuestionHelper.php(63): Symfony\\Component\\Console\\Helper\\QuestionHelper->doAsk(Object(Illuminate\\Console\\OutputStyle), Object(Symfony\\Component\\Console\\Question\\ConfirmationQuestion))
#2 /var/www/attendize/vendor/symfony/console/Style/SymfonyStyle.php(294): Symfony\\Component\\Console\\Helper\\QuestionHelper->ask(Object(Symfony\\Component\\Console\\Input\\ArrayInput), Object(Illuminate\\Console\\OutputStyle), Object(Symfony\\Component\\Console\\Question\\ConfirmationQuestion))
#3 /var/www/attendize/vendor/symfony/console/Style/SymfonyStyle.php(222): Symfony\\Component\\Console\\Style\\SymfonyStyle->askQuestion(Object(Symfony\\Component\\Console\\Question\\ConfirmationQuestion))
#4 /var/www/attendize/vendor/laravel/framework/src/Illuminate/Console/Command.php(314): Symfony\\Component\\Console\\Style\\SymfonyStyle->confirm('Do you really w...', false)
#5 /var/www/attendize/vendor/laravel/framework/src/Illuminate/Console/ConfirmableTrait.php(31): Illuminate\\Console\\Command->confirm('Do you really w...')
[...]

PR #552, opened 2019-01-12
Force key generate in installer

delenamalan:
Addresses #458 (and related #524). Forces the application key to be generated.

The `STDIN` error is caused by artisan which expects an input:
```
$ php artisan key:generate
**************************************
*     Application In Production!     *
**************************************

 Do you really wish to run this command? (yes/no) [no]:
 > 
```

We can bypass having to enter an input by forcing the key to be generated.


jeremyquinton:
@delenamalan nice fix this I struggled to replicate this one.

PR #558, opened 2019-01-16
Hidden ticket access codes (#1)

etiennemarais:
# Summary

Added the ability to specify hidden ticket access codes on an event. This would allow for special types of tickets to be created on an event and have an unlock code link to one or many hidden tickets.

## Value

The value this adds is to allow event organisers to play with marketing to aid in ticket sales etc.

## Changes

**Admin**
- Added migrations to allow for the access codes table on events and the pivot for many to many codes to tickets
- Added the ability to create an access code in the event customisation 
- Added the ability to view access codes on a hidden ticket
- Added the ability to attach one or multiple access code(s) onto the hidden ticket

**Public event page**
- Shows a box to enter your access code if the event has hidden tickets
- Added the validation messages for empty codes
- Added the check to fetch the hidden tickets if the code was entered correctly


modifiedcontent:
Is there a way to make the 'Do you have an unlock code?' option show up in the widget, with the rest of the registration form? I only get it in the event page, which I don't use.

samdb:
We haven't done any work on the widgets for a while as they get blocked by some popular browsers (Safari for example - see issue #607) when trying to submit the form. I am looking at building a replacement widget set that can cater for this.

modifiedcontent:
Thank you for your response.

Is there a quick way to get it in the widget? Which file(s) should I edit?

I know iframe is problematic. It would be nice if I could just php include the form on another page on another domain, but on the same server.

PV

⁣

PR #560, opened 2019-01-17
fix broken images

dillingham:


jeremyquinton:
@dillingham thanks for this.

PR #569, opened 2019-02-09
Add a single organiser mode

publicarray:
Useful to prevent the public from creating Events/Organisations without authorisation. (should be enabled by default IMHO; but that would break compatibility) 

I would use this where there is only one organisation publishing events and where we don't want to allow the public to create Events/Organisations

(I think this should be enabled by default IMHO; but that would break compatibility)


Bellusterra:
@publicarray I actually can’t figure out how to allow the public to create events :). It appears that the admin can create users and add organizers, but I don’t see a public facing registeration page.

publicarray:
The public can create new organisation admin accounts and organisers from `/signup` from there users within the organisation can create new events

Bellusterra:
Thank you @publicarray, I must have missed that... I agree the ability to disable that function is a great option to have.

etiennemarais:
@publicarray Thank you for contribution. I am loojing to get this merged into our `next`  branch. Please help with the conflicts? 🙏 The `next` branch has recently undergone a large laravel version upgrade _(6.5.1)_ Please use the new routes file since it has moved

publicarray:
Closing in favour of new pull request on the updated branch `next` #741 🤩

PR #571, opened 2019-02-13
Include log Facade

vikashsp:


jeremyquinton:
@vikashsp can you explain the reason for this change?

vikashsp:
> @vikashsp can you explain the reason for this change?

We're using this log facade without loading it int the controller, it gives an error.
Check below line:
https://github.com/Attendize/Attendize/blob/810f40c1e3ee01ac03bef2cc03f8bd237bbdf632/app/Jobs/SendOrderNotification.php#L38
https://github.com/Attendize/Attendize/blob/810f40c1e3ee01ac03bef2cc03f8bd237bbdf632/app/Jobs/SendOrderNotification.php#L39
https://github.com/Attendize/Attendize/blob/810f40c1e3ee01ac03bef2cc03f8bd237bbdf632/app/Jobs/SendOrderNotification.php#L40

PR #572, opened 2019-02-13
Feature/Discount code delete and usage count

etiennemarais:
# Summary

We needed a way to soft delete the event access codes and track their usage count. The delete needed to happen only if the code hasn't been used before and we needed to check if the event actually had the code linked to it. We also needed to move the access codes to the event side nav below the `Event Tools` section as it's own page.

## Changes

- Moved the access codes code and templates to the side nav
- Added a migration for usage count that defaults to zero
- Added the ability to remove an access code on an event only if it hasn't been used before
- Moved the access code translations to it's own file called `AccessCodes`
- Moved the Access code controller code to it's own controller
- Moved (#568) here. This is the check against adding duplicate access codes

## Screenshots

screen shot 2019-02-14 at 17 12 52
screen shot 2019-02-14 at 17 12 48


Iamscalla:
To be fair, there is a difference between access code and discount codes.. So renaming it is not wise. Discount code stands on it own.

etiennemarais:
Yes I do agree. @samdb, what are your thoughts? @Iamscalla is correct in that a discount code is generally accepted as a `10%` off or a one time `40GBP` off a ticket price whereas these access codes are given at discretion to show hidden (cheaper/other) tickets on a specific event. Renaming is easy though.

The specification for this PR required me to rename it and move it to the sidebar though for extra context @Iamscalla. 

Thoughts @jeremyquinton ?

samdb:
Ok we can leave it as access codes until a proper discount code feature is in place. Once that is done is can be a single section called Discount and Access Codes. This is what other platforms seem to do:

screenshot 2019-02-14 at 10 45 29

screenshot 2019-02-14 at 10 45 58

jeremyquinton:
@etiennemarais I have tested this and it is working well. 

My only feedback is a small UI bit. When you click hide ticket you first have to save the ticket then edit it again before you can apply an access code to it. Not a train smash but we can add that improvement at later stage. 

Going to merge this.

PR #581, opened 2019-02-25
Japanese translation

matometaru:
I will create a Japanese translation file.


jeremyquinton:
@matometaru thanks for this. When this is complete I will be happy to merge it.

matometaru:
Translation is complete.
I have not seen all the pages yet, so there may be strange translations.
If there is a strange translation, we will resubmit with another pull request.

Bellusterra:
I had just finished translating all the strings, but used google translate, and was going to have some help with adjusting from there. So this is great, thank you @matometaru

PR #585, opened 2019-02-28
Add reCAPTCHA

publicarray:
Adding this can help prevent simple bots from using credential stuffing to guess multiple passwords. I'm not sure if Laravel's rate-limiting is used, even if it is this can add additional layer of protection.


Iamscalla:
i felt there have to be a condition attached before captcha appears. 
conditions like number of failed attempt or changed IP

publicarray:
This uses v3 of recaptcha which does not challenge the user. It returns a score for each request [anything lower than 0.6 is blocked](https://github.com/Attendize/Attendize/pull/585/files#diff-7d19adb96239741ff7ecbf7658c1766cR77)

https://github.com/Attendize/Attendize/blob/6dff99b9bfef6faaca8f60ce73f801d6a67afffd/app/Http/Controllers/UserLoginController.php#L77

(1.0 is very likely a good interaction, 0.0 is very likely a bot). Since it's mostly admins that need to login I choose a higher threshold.

Iamscalla:
screenshot_2019-02-28 nexmo - apis for sms voice and phone verifications 1


@publicarray you mean as seen in the image above??

jeremyquinton:
@publicarray in principle this is a good idea would be happy to merge. See my comments above. Additionally what @Iamscalla suggests about only showing a captcha for failed login attempts is not a bad idea.

samdb:
I would still like to merge this PR, though I will likely make $data->score another config variable.

PR #600, opened 2019-03-12
manage event currency

jdnwi:
Event currencies were previously only saved upon event creation and weren't able to be updated.  You can now change currency at the event level.


jeremyquinton:
@jdnwi thanks for this. This is very useful to have in. The only confusion is that the default currency can be set at the organiser level. On the organiser dashboard it shows total sales in once currency. I will update that dashboard to show the sales total in different currencies.

PR #612, opened 2019-04-03
Update datepicker to fix disapearing datepicker in chrome

mrlexley:


jeremyquinton:
@mrlexley thanks for this fix much appreciated

PR #616, opened 2019-04-05
Change Private Reference Number

mrlexley:
This changes the Private Reference Number (Data in QR Code), less chance of getting double ID's and I think its more secure.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,6 +11,7 @@ installed
 /composer.phar
 _ide_helper.php
 *.swp
+composer.lock
 
 # Do not include backup lang files
 resources/lang/*/[a-zA-Z]*20[0-9][0-9][0-1][0-9][0-3][0-9]_[0-2][0-9][0-5][0-9][0-5][0-9].php
--- a/.gitignore
+++ b/.gitignore
@@ -11,7 +11,7 @@ installed
 /composer.phar
 _ide_helper.php
 *.swp
-composer.lock
+
 
 # Do not include backup lang files
 resources/lang/*/[a-zA-Z]*20[0-9][0-9][0-1][0-9][0-3][0-9]_[0-2][0-9][0-5][0-9][0-5][0-9].php
--- a/app/Models/Attendee.php
+++ b/app/Models/Attendee.php
@@ -45,12 +45,12 @@ class Attendee extends MyBaseModel
         parent::boot();
 
         static::creating(function ($order) {
-			
+
             do {
                 //generate a random string using Laravel's str_random helper
-                $token = Str::Random(10);
+                $token = Str::Random(15);
             } //check if the token already exists and if it does, try again
-			
+
             while (Attendee::where('private_reference_number', $token)->first());
             $order->private_reference_number = $token;
         });
--- a/app/Models/Attendee.php
+++ b/app/Models/Attendee.php
@@ -45,8 +45,14 @@ class Attendee extends MyBaseModel
         parent::boot();
 
         static::creating(function ($order) {
-            $order->private_reference_number = \Str::Random(10);
+            do {
+            //generate a random string using Laravel's str_random helper
+            $token = Str::Random(10);
+            } //check if the token already exists and if it does, try again
+            while (Attendee::where('private_reference_number', $token)->first());
+            $order->private_reference_number = $token;
         });
+
     }
 
     /**
--- a/app/Models/Attendee.php
+++ b/app/Models/Attendee.php
@@ -45,10 +45,12 @@ class Attendee extends MyBaseModel
         parent::boot();
 
         static::creating(function ($order) {
+			
             do {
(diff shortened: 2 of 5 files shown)


jeremyquinton:
@mrlexley I also noticed a few months ago its possibly to get potentially classing ids. You mention less chance in your PR is there a way you suggest we can make it always unique and still be random?

mrlexley:
The chance is less with alpha numeric. But there is still a very very low change of getting a duplicate ID. But the risk is very small. Its more about fake tickets that people possibly can generate and come to an event. That's why I made the change in this pull. If you want to full cover the duplicate ID part the best would be if the ID not exists else create a new one. But the same needs to happen for the order IDs if I'm not wrong.

jeremyquinton:
@mrlexley thanks for your reply you have made some very valid points. I'd need to give this a bit more thought but will put a fix in place in the coming weeks as potentially duplicate ids is no good.

mrlexley:
@jeremyquinton No, it isn't. But this way the chance of getting duplicates is very small there are 7^62 = 35.216.146.062.080 combinations possible instead of 999.999.999. But it's better to let the model check if I'd exist ofcource. Also now I think of it. I only tested it with qr codes not the 2d barcode.

mrlexley:
https://laracasts.com/discuss/channels/tips/generate-a-unique-random-string-for-the-database

.gitignore
```diff
@@ -11,6 +11,7 @@ installed
 /composer.phar
 _ide_helper.php
 *.swp
+composer.lock
```
publicarray: Why add composer.lock to gitignore? I like to keep it in git, so that there is always a lit of dependencies that are known to work. You can call `composer update` on your own fork/install whenever you want.
https://getcomposer.org/doc/01-basic-usage.md#commit-your-composer-lock-file-to-version-control

app/Models/Attendee.php
```diff
@@ -44,7 +45,7 @@ public static function boot()
         parent::boot();
 
         static::creating(function ($order) {
-            $order->private_reference_number = str_pad(random_int(0, pow(10, 9) - 1), 9, '0', STR_PAD_LEFT);
+            $order->private_reference_number = \Str::Random(10);
```
etiennemarais: I would suggest making this the same length as the database field for brevity sake. Also lessen the chance of collisions. 

```
\Str::Random(15); // The same length as the migration
```

database/migrations/2019_04_05_180853_change_private_reference_number_column_type.php
```diff
@@ -0,0 +1,32 @@
...
+class ChangePrivateReferenceNumberColumnType extends Migration
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('attendees', function ($t) {
```
etiennemarais: A small nitpick but please add the type indicator to the closure of both `up()` and `down()` methods? 

```
function (Blueprint $table) {
```

and use the full variable name for the migration in the `up()` and `down()` calls please? 🙏

.gitignore
etiennemarais: I feel this needs to be a separate PR as to discuss the reasons why we need to ignore or not ignore this file. Please undo this change from this PR for the time being? 🙏

mrlexley:
Changes are made, I also added a check if order_reference already and private_reference_number already exists else create new one. Removed composer.lock from gitignore.

etiennemarais:
Thank you, Please align your do/while loops and update the migration files as well please?

mrlexley:
What do you mean by making the migration files. Migrations are already made.

mrlexley:
> 1. Please update the migration files to indicate type hints: `function (Blueprint $table) {` on both
> 2. Please use the full length `Str::random(15)` to utilise the full length of the db field

Made changes in migration, can you check if this is oke?

mrlexley:
Forgot to push changes, but you can check now.

etiennemarais:
> Forgot to push changes, but you can check now.

Thank you, The migration changes are perfect thanks. 

> Please use the full length Str::random(15) to utilise the full length of the db field

This part is still not finished unfortunately. Because the database field is `15` characters in length, Please can you change the instances of `Str::random()` to also be `15` in length? 🙏 This will ensure we utilise the space provided but also make sure there is a smaller chance of collisions.

mrlexley:
> > Forgot to push changes, but you can check now.
> 
> Thank you, The migration changes are perfect thanks.
> 
> > Please use the full length Str::random(15) to utilise the full length of the db field
> 
> This part is still not finished unfortunately. Because the database field is `15` characters in length, Please can you change the instances of `Str::random()` to also be `15` in length? 🙏 This will ensure we utilise the space provided but also make sure there is a smaller chance of collisions.

Made the change, just used the same field length and string length as reference number before. But I made it 15 now.

PR #626, opened 2019-04-25
Business fields for tax purposes

etiennemarais:
# Summary

We needed a way to enable businesses to show that they are tax registered and add that to the ticket details, receipts and email to allow them to claim back as a business expense.

## Changes

- Added business details fields migration to the `orders` table
- Added the input fields on the Register screen
- Added the functionality to allow toggle for business
- Show the business details on the email
- Show the business details on the confirmation screen
- Added translations for the new labels
- Added a package + helper to deal with cleaning the input better
- Hide the booking fee if it is zero


publicarray:
This is a great PR!

Regarding sanitisation, I see what you are trying to do but I would prefer not complicate the situation further by adding yet another function that doesn't protect all inputs. In an ideal world we would only use the `{{` `}}` tags to print user content / untrusted content from the database to html. Again personal opinion, but I also don't like the `{!! Form::text` stuff either, why not use good old html? To be clear these points aren't directed at you just my position about it and this project. Maybe one day I get around fixing it in a PR. Basically this whole thing is a bit of a mess. e.g. https://github.com/Attendize/Attendize/pull/629

https://github.com/OWASP/CheatSheetSeries/blob/master/cheatsheets/Cross_Site_Scripting_Prevention_Cheat_Sheet.md

jeremyquinton:
@etiennemarais have tested this PR and it works.

One small issue is. 

The email that gets sent to the person booking the tickets should contain the business details. 
Currently the email that is sent to the organiser contains the business details. 
Can you update the email template that gets sent to the person booking the tickets.

@samdb question for you. 
Once this is merged someone can book tickets but not enter their business details. At a later stage they might want to add business details to their booking. Should they be able edit an order and add business details at a later stage.

etiennemarais:
@jeremyquinton I made the update to the tickets email. 🙏 review?

robetus:
Is there a way to disable the toggle on the register page?

joleys:
> 

@robetus Did you get an answer to your question?

PR #628, opened 2019-04-27
Bug : total sales is incorrect (organiser fees are included twice)

loup-brun:
Hi, I was trying to figure out the problem with my `sales volumes`.

It turns out that if I set a fee for my event, **the fee is already included in the event's total sales**, which means adding fees again produces inflated numbers.

This is apparent in the dashboard totals.

I fixed the 2 commits in my installation, hence the PR :smile:

### DIFF
--- a/resources/views/ManageEvent/Dashboard.blade.php
+++ b/resources/views/ManageEvent/Dashboard.blade.php
@@ -100,8 +100,13 @@
                         <div class="panel-heading panel-default">
                             <h3 class="panel-title">
                                 @lang("Dashboard.ticket_sales_volume")
+<<<<<<< HEAD
                                 <span style="color: green; float: right;">
                                     {{money($event->sales_volume + $event->organiser_fees_volume, $event->currency)}}
+=======
+                                <span class="panel-meta-success">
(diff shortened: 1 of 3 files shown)


resources/views/ManageEvent/Dashboard.blade.php
```diff
@@ -100,8 +100,13 @@
                         <div class="panel-heading panel-default">
                             <h3 class="panel-title">
                                 @lang("Dashboard.ticket_sales_volume")
+<<<<<<< HEAD
```
publicarray: By the looks of things you had a merge conflict and you haven't resolved it.

I think the easiest way to fix this is by a adding a new commit to remove the `<<<<<<< HEAD`, `=======` and `>>>>>>> dcd171a... Correct sales volume in 2 other files` lines.

loup-brun: Sorry about that – fixed in 3e13e5a77cee8e6b072fdeb44ef18389eac8681f

etiennemarais:
This has been fixed in #721

PR #630, opened 2019-04-30
Remove closures on route files

publicarray:
* `php artisan route:cache` now works 🚀 
* Fixed 1 Todo
* Removed the sample API route `/`


jeremyquinton:
Hi @publicarray.

Thanks for this PR it is a good improvement. 

when I run ```php artisan route:cache``` I still get an error about the terms and conditions. I'll fix that one. 

In Route.php line 917:
                                                                                   
  Unable to prepare route [terms_and_conditions] for serialization. Uses Closure.

PR #631, opened 2019-05-03
Add price to options

publicarray:
This gives us the ability to charge for certain extras such as food or drinks.


publicarray:
If you like I can squash the commits down.

justynpride:
I love this feature. One suggestion for future improvement is for when the select option doesn't have or need a price. It would be great if the price field could not be displayed in those situations when booking. An alternative is to have a flag on the survey question to say whether the price option should be displayed or not. 
Also to be able to control the quantity of the items available would be a huge bonus in some situations way numbers are limited (ie a seminar and room capacity).

justynpride:
I've merged this branch to test further, but I can't see to get the ticket prices to increase based upon any add on options I've added.

publicarray:
Do you mean the right hand side on the checkout form, yea thanks for reminding me I totally forget about that. 🙊

justynpride:
Yes the right hand side. Glad to have been of help.

publicarray:
@justynpride The price should now update on the checkout page. 👍

justynpride:
Looks good on the check outside in terms of the calculation. I've having problems checking out, but I think that is my install. 

One comment thought: if I add any of those survey questions but don't have a price for them, they now show up in the checkout, which they didn't before. A simple example would be if I asked for Gender. It's not something I'd want to show up in the checkout, but it does. Is the answer a flag on the question to ask whether to display in checkout? 

From my personal perspective my ideal displaying of the details would be for them to be a subsection of the ticket, so for a ticket called Adult, with a meal plan option of '2 meals', and another option of merchandize 'Mug' it would show Adult - 2 Meals, Mug. I don't think would be readily doable in Attendize though as tickets are totalled together, rather than being individually listed.

justynpride:
@publicarray That's perfect, they don't show up now. Great work. I can think of a number of uses for this feature.

g2petter:
Has this feature been made available in the master branch? I'm testing out the demo website and I can't find any way to enable it, at least.

samdb:
We’re not very good at keeping the demo site up to date. It is probably very old.

I will look at setting up some kind of CI from the master branch.

justynpride:
It's never been merged into the master.

g2petter:
@justynpride ah, my bad. 

Do you know if you have any plans to accept this pull request?

g2petter:
Ah, my bad.

@samdb do you know if you have any plans to merge this pull request?

justynpride:
I agree with @g2petter, would love to know if this is likely to be added into the new version now that it has been released?

samdb:
Now that we have a better foundation with the latest release we will be more comfortable testing and merging PRs. This one is on the list for review.

justynpride:
@publicarray had you this working on the latest release? I know it’s been a over a year since the original pull request.

publicarray:
Hi @justynpride,

I'm Sorry to say that I've not looked at the code since and as you said it's been a long time. Now there are no large number of conflicts. Sorry I know it's not the answer you where looking for but maybe someone can take the general idea and re-implement it.

justynpride:
Hi @publicarray 
As always thanks for the helpful reply. I had wondered whether that might be case, but always like to ask. I'll add it to a 'to look into' list. I've a few higher priorities for possible dev, but being able to have 'add ons' for tickets is still quite high.

justynpride:
> Hi @publicarray
> As always thanks for the helpful reply. I had wondered whether that might be case, but always like to ask. I'll add it to a 'to look into' list. I've a few higher priorities for possible dev, but being able to have 'add ons' for tickets is still quite high.

From what I can see and test it falls down on the final part of the checkout after tickets and the correct fee has been calculated. The order checkout doesn't pickup on the new total including extras, and there are problems with the payment gateway. Thought it worth feeding back here in case anyone has the gift and time to take this over the finishing line.

PR #632, opened 2019-05-04
fix(qr-code scanning)

### LINKED ISSUE #614: Which device use for QR Code and How To Setup In Local Device
Hi,

Thanks for making this type of Good System I like it..

I have a question , How i read QR Code in below screen(sent screenshot) , I tried my system (laptop) camera but that camera not able to read this so Is there any device I need to purchase or any way for read this QR Code ?

image

Please help me

Thanks&Regards

mstaack:
this works for me with ios 12.2 and safari. chrome on iphone does not work. 
not familiar with `getUserMedia`, but at least one browser works now.

### DIFF
--- a/public/assets/javascript/check_in.js
+++ b/public/assets/javascript/check_in.js
@@ -107,9 +107,27 @@ var checkinApp = new Vue({
             }
 
             qrcode.callback = this.QrCheckin;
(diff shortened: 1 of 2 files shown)


gerogr:
doesn't work for me with Chrome, Vivaldi on Macbook latest version and Chrome on Android

PR #636, opened 2019-05-06
Move env to config files

publicarray:
This fixes missing environment values when using `php artisan config:cache` as this should be done in production.


jeremyquinton:
@publicarray thanks for this improvement.

PR #640, opened 2019-05-14
Add GTM container

etiennemarais:
## Summary

We needed to add google tag manager to the public event pages to allow someone to track sales among other things on their live events.

Basic logic here is that the GTM code will live on both the `event` and `organiser` and will cascade in rendering in the following order:

1. If the `event` has a gtm code, it will render it first
2. If the event does not have a code, the `organiser` code will be used
3. Will not render the GTM scripts if both don't have codes set

## Changes

- Added the migrations to add a field on both the `event` and `organiser`
- Added the text fields to allow a user of Attendize to update the `organiser` and/or `event` with the GTM code
- Updated the public event page to render the GTM script in the `<head>` if a valid code exists


resources/views/Public/ViewEvent/Partials/GoogleTagManager.blade.php
```diff
@@ -0,0 +1,18 @@
+<?php
+$gtm_code = '';
+if (empty($event->google_tag_manager_code) === false) {
```
saiht:
```suggestion
if (!empty($event->google_tag_manager_code)) {
```
suffices

saiht: same below

public/assets/javascript/backend.js
```diff
@@ -6422,38 +6422,6 @@ if(!Object.keys)
     };
 }
 
-// =================================== START DATEPICKER =====================================
```
saiht: all cleanup here, public/assets/stylesheet/application.css and app/Http/Controllers/OrganiserCustomizeController.php should be in a separated commit / PR, what do you think ?

PR #642, opened 2019-05-21
Fix French translation

bagage:


samdb:
Hi @bagage, the wording is meant to suggest that the question is mandatory. Does this new translation make that point?

bagage:
The point is that this is actually not mandatory.
It only enables the question for givin ticket, but it does not make it mandatory.

PR #645, opened 2019-05-28
Calculations fixes

etiennemarais:
## Summary

The event calculations needed some love and checked against the use cases. There were problems when refunding people, *not with the refunding part*, but with the event and ticket revenue stats in some cases doubly decrementing the event and not showing the ticket revenue less tax/refunds.

### Refunds

Refunds are now linked to a single or list of attendees. A refund will cancel the attendee or list of attendees and flag each one as refunded. A refund cannot happen without a subsequent cancellation anymore.

Orders will now get marked as partially refunded when a subset of attendees are cancelled/refunded on a given order. Tax will be refunded and calculated on a per ticket basis using the ticket price and the organiser tax amount if it's set. Basically each refund will contain the tax amount refunded per attendee.

### Partial Refunds

Partial refunds have been removed. An order can still be marked as a partial refund but the concept of partially refunding an order without cancelling the attendees have been removed because it breaks the refund and stats counting flows in the system.

This ties into the above mentioned point that a refund must be linked to an attendee cancellation.

### Retrofit amounts and stats

This PR also contains a migration fix script that will retroactively goes back in time and attempt to fix the `tickets`, `event_stats`, `orders` and `attendee` tables to reflect the current order and order items information. You can view the log debug output to see which records were affected.

- Linked tickets to their orders
- Fixed orders with missing values
- Fixed ticket sales volume and quantity sold stats based on orders and order items
- Fixed event time series stats based on orders and order items
- Fixed attendee status flags for refunded / cancelled
- Fixed broken event stats entries by clearing the tickets sold and sales volume values if there is no corresponding order information to back them. Kept the views and unique views though
- Added a local test seeder to aid in setup when developing locally
   - First building block to allow for feature tests to become a thing

### Changes

- Added validation to the refund section by requiring attendees to be set when cancelling an order.
- Removed the event `sales_volume` and `organiser_booking_fees` fields on the `event` table in favour of the `event_stats` table.
- Added helper fields to calculate and use the correct revenue fields for a ticket and event.
- Fixed an issue where form validation errors would not have a place to render the error for inputs that are an array type, ex: `attendees[]`
- Did some general spacing, alignment and cleansing boy scout things.
- Added a Money composer lib that uses `bcmath` underneath to help calculate the various floats and compare them accurately.
- Fixed the issue where the tickets were not getting logged against an order properly. This is a many to many pivot table that never seemed to work. I added the fixes for the ticket to be listed against an order to aid in proper revenue calculations. The big error here was a copy paste issue where the foreign key index on the `ticket_id` was against the `users` table 😕 
- Fixed the order amount getting removed on a refund. This completely breaks the partial refund calculations, which is now fixed.
- Fixed the order manage modal receipt to show the full amounts of the order but indicate if the order amounts were refunded along with the `cancelled`/`refunded` tags now showing on the attendee if they have been refunded and cancelled.
- Added helper amount methods to the order model to allow for easy displays of things like `MaximumRefundableAmount` etc.
- Fixed the order cancellation screen to cancel and refund an attendee or list of attendees from the selection.
- Fixed the single attendee cancellation by forcing the refund to happen instead of making it optional. This includes calculating tax if it's set on the organiser.
- General UI updates to simplify the refund/cancellation flow.


jeremyquinton:
@etiennemarais I have done some testing. Found some weird bugs.

I checked out this branch locally and made a completely fresh install. 

1. Create event
If you leave the description empty but try and save it there is no error message displayed to the user. 
Confirmed this works on Master. 
2. Create tickets with a cost of £100.   
3. Purchases 1 tickets.
4. Go the the event dashboard. Ticket sales volume is £0.00 
5. Go to orders and partially refund the ticket £90. 
6. Go to the event dashboard the revenue is £0. Should it not be £10.
7. Go to orders and try refund the last remaining £10 by issuing a full refund or issuing a partial refund. Json response. 
{status: "error", messages: {attendees: ["Please select at least one attendee"]}}
    1. messages: {attendees: ["Please select at least one attendee"]}
        1. attendees: ["Please select at least one attendee"]
            1. 0: "Please select at least one attendee"
    2. status: "error"
I think its because the the attendee is cancelled. Error is also not displayed to the user. 

Over and above this I took a copy of the production database which has a couple events setup. I imported the database and then ran the migrations on your branch. On the events dashboard the revenue for all the events is suddenly wrong. If you want I can provide you with a copy of this database I will just anonymise the data.  

I'm not an expert on Eloquent in Laravel however as you have mentioned above 
```Fixed the issue where the tickets were not getting logged against an order properly. This is a many to many pivot table that never seemed to work. I added the fixes for the ticket to be listed against an order to aid in proper revenue calculations. The big error here was a copy paste issue where the foreign key index on the ticket_id was against the users table```

I think as part of this work we should potentially correct the database model. 
The relationships are as follows in my head but I might miss something. 

1. 1 Order can have many tickets. 1 to many. 
    I think the tickets should be like line items on an order.
2. 1 Ticket can have 1 attendee associated with it.

Then

1. Orders can be refunded in full. This would be a line item on the order.
2. Orders can be refunded partially. This would also be a line item but it would be associated with a ticket which is refunded. Need to decide if we partially refund a particular ticket e.g ticket price was £150 but the attendee is refunded £130. 

The way this is modelled currently is not great I remember seeing is_refunded, is_cancelled flags on attendees and orders and refunded fields on tables. The reality is making these changes might mean that users can't upgrade their codebase but we could potentially write a migrate tool to migrate to the new schema.

**Comment by jeremyquinton** [comment]
Elaborating on what I said above. An example order below without a refund. Currently refunds go against the order. This creates ambiguity around partial refunds. If something is partially refund do we cancel the order and the attendee. 

mysql> select * from orders\G
*************************** 1. row ***************************
                   id: 1
           account_id: 1
      order_status_id: 1
           created_at: 2019-05-31 15:21:17
           updated_at: 2019-05-31 15:21:17
           deleted_at: NULL
           first_name: Jeremy
            last_name: Quinton
                email: jeremyquinton@gmail.com
      ticket_pdf_path: NULL
      order_reference: PDUAQ315
       transaction_id: ch_1EgCXhIyUrX5heus7OmOePg2
             discount: 0.00
          booking_fee: 0.00
organiser_booking_fee: 0.00
           order_date: NULL
                notes: NULL
           is_deleted: 0
         is_cancelled: 0
is_partially_refunded: 0
          is_refunded: 0
               amount: 100.00
      amount_refunded: NULL
             event_id: 1
   payment_gateway_id: 1
  is_payment_received: 1
               taxamt: 0.00
1 row in set (0.00 sec)

mysql> select * from orders_items\G
ERROR 1146 (42S02): Table 'attendize.orders_items' doesn't exist
mysql> select * from order_items\G
*************************** 1. row ***************************
              id: 1
           title: General Admission
        quantity: 1
      unit_price: 100.00
unit_booking_fee: 0.00
        order_id: 1
      deleted_at: NULL
1 row in set (0.00 sec)

mysql> select * from attendees\G
*************************** 1. row ***************************
                      id: 1
                order_id: 1
                event_id: 1
               ticket_id: 1
              first_name: Jeremy
               last_name: Quinton
                   email: jeremyquinton@gmail.com
private_reference_number: EhonvQc9gvhANzD
              created_at: 2019-05-31 15:21:17
              updated_at: 2019-05-31 15:21:17
              deleted_at: NULL
            is_cancelled: 0
             has_arrived: 0
            arrival_time: NULL
              account_id: 1
         reference_index: 1
             is_refunded: 0
1 row in set (0.00 sec)

When something is partially refunded how does it affected the order as a whole. I think for now if you fixed the bugs mentioned in the first part of the PR we could most likely still merge it.

**Comment by jeremyquinton** [comment]
Some thoughts I put together a while ago which might also help. 

https://docs.google.com/document/d/1S-yvcQqWl7se1UxkMXzV4nmTxiYfrd6kWLXFZoqnLfs/

**Comment by etiennemarais** [comment]
@jeremyquinton 

You can now run this test seed to aid in testing locally faster.

`php artisan migrate:refresh --seed && php artisan db:seed --class=LocalTestSeeder`

and you can log in with `local@test.com:pass`.

### Noticeable changes

- I've removed the notion of refunding custom amounts
- I've fixed the process of refunding to tie an attendee cancellation to a refund so the two can't happen independently from one another anymore skewing the stats
- Local database test seed to aid in testing changes added but removed from main migrations
- Fixed graph totals on the dashboards
- Partial refund status still remains as a way of indicating 1 or more attendees from an order was refunded but not all

File: app/Http/Controllers/EventOrdersController.php [hunk 1]
```diff
@@ -220,112 +215,131 @@ public function postCancelOrder(Request $request, $order_id)
...
+        // Get the full order amount, tax and booking fees included
+        $organiserAmount = new Money($order->organiser_amount, $order->getEventCurrency());
+        Log::debug(sprintf("Total Order Value: %s", $organiserAmount->display()));
+        $refundedAmount = new Money($order->amount_refunded, $order->getEventCurrency());
+        Log::debug(sprintf("Already refunded amount: %s", $refundedAmount->display()));
+        $maximumRefundableAmount = $organiserAmount->subtract($refundedAmount);
+        Log::debug(sprintf("Maxmimum refundable amount: %s", $maximumRefundableAmount->display()));
+        /*
+         * NOTE: Calculates the refund amount by using the order total (tax incl) and dividing it by the amount of
+         * attendees the user has selected. This will evently spread the tax refund per attendee.
```
jeremyquinton: small typo in the spelling of evenly

app/Http/Controllers/EventOrdersController.php
```diff
@@ -220,112 +215,131 @@ public function postCancelOrder(Request $request, $order_id)
...
+                'transactionReference' => $order->transaction_id,
+                'amount' => $refundAmount->toFloat(),
+                'refundApplicationFee' => floatval($order->booking_fee) > 0 ? true : false,
+            ]);
+
+            Log::debug("STRIPE REQUEST", [
+                'transactionReference' => $order->transaction_id,
+                'amount' => $refundAmount->toFloat(),
+                'refundApplicationFee' => floatval($order->booking_fee) > 0 ? true : false,
+            ]);
```
jeremyquinton: Hard coding STRIPE REQUEST here is not good as Attendize uses Omnipay so probably better to log the gateway_name

resources/views/ManageEvent/Orders.blade.php
```diff
@@ -114,12 +114,10 @@
                                 > {{$order->email}}</a>
                             </td>
                             <td>
-                                <a href="#" class="hint--top" data-hint="{{money($order->amount, $event->currency)}} + {{money($order->organiser_booking_fee, $event->currency)}} @lang("Order.organiser_booking_fees")">
-                                    {{money($order->amount + $order->organiser_booking_fee + $order->taxamt, $event->currency)}}
-                                    @if($order->is_refunded || $order->is_partially_refunded)
-
-                                    @endif
-                                </a>
+                                <span>{{ $order->getOrderAmount()->display() }}</span>
```
jeremyquinton: The back-end orders dashboard the order total is displayed EX V.A.T is this intentional?

I think the getOrderAmount method needs to include the tax amount on the order.

app/Http/Controllers/EventOrdersController.php
```diff
@@ -220,112 +215,131 @@ public function postCancelOrder(Request $request, $order_id)
...
+        Log::debug(sprintf("Maxmimum refundable amount: %s", $maximumRefundableAmount->display()));
+        /*
+         * NOTE: Calculates the refund amount by using the order total (tax incl) and dividing it by the amount of
+         * attendees the user has selected. This will evently spread the tax refund per attendee.
+         *
+         * If only one attendee lives in an order then its safe to assume a full refund can be made.
+         */
+
+        $refundAmount = $maximumRefundableAmount->multiply($attendees->count())
+            ->divide($nonCancelledOrderAttendees->count())->floor();
```
jeremyquinton: Lets take the following scenario

Ticket 1 £400
Ticket 2 £400
Ticket 3 £150

Total order ex vat is £950. Total order with V.A.T is £1140.00.

I refund a single ticket of £400. 

The above calculation results in £380 however the ticket price is £480. The above calculation will work refunding the order as a whole but I think that’s quite confusing if the intention is to refund one ticket. 

I think this will need to change to refund the individual ticket as a whole

resources/views/ManageEvent/Orders.blade.php
etiennemarais: @samdb mentioned that all admin side amounts should exclude tax except the tax entries on the order details.

etiennemarais:
```
Refunding
If an order is awaiting payment and you try to refund cancel it get a 500 error.

Revenue Total
The event revenue total is coming out ex vat its not including the V.A.T amount from the order table.
```

1. I will have a look at the `500` happening here. It should be a simple issue
2. The revenue total should exclude tax for all amounts inside of the admin area as part of the requirements for this change as tax is not revenue.

app/Http/Controllers/EventOrdersController.php
etiennemarais: Busy looking at this as well. Thanks for reporting 🙏

etiennemarais:
To fix:

- [ ] Organiser tickets sold does not subtract cancelled attendees

etiennemarais:
@jeremyquinton I fixed both those tested use cases we found.

CristianDeluxe:
This is gonna be awesome! Wishing that we could include many more things in Attendize in a simpler way.

Above all: the test part, very necessary.

But I also love refactoring =) I would love Attendize controllers to be much lighter than they are today.

PR #652, opened 2019-06-17
Update to Laravel 5.8

CristianDeluxe:
Finally from PR #540 I have separated only the update to Laravel 5.8

I know that there are many files changed but it is necessary to do it well and I think that a "step by step" update would give many errors.

--- Some parts of text from the original PR ---

I created a new Laravel project and copied file by file from Attendize, integrating, updating and merging.

It's a big change (a lot of files) but I think it was necessary to start as clean as possible.

You can follow the changes more simply by looking at my commits separately.

- Updated all to 5.8 and tested
- Updated laracasts/utilities to v3
- Now all routes use the new Laravel 5.8 syntax, which has allowed to discover an error in the namespace of API routes.
- The helpers have been rewritten and now use ServiceProviders
- Deleted some Service Providers and Aliases from config/app.php file because from Laravel 5.5 some have activated the auto discovery function.
- Fixed some tests, although some functions are not available, I think it is not worth spending more time as they need to be rewritten from scratch.

### Breaking Changes
- Now by default is used utf8mb4 encoding that supports emojis and comes with Laravel 5.8
- The images in the events table should be renamed from assets/images/.. to images/...

### DIFF
--- a/app/Services/PaymentGateway/Factory.php
+++ b/app/Services/PaymentGateway/Factory.php
@@ -2,11 +2,7 @@
 
 namespace App\Services\PaymentGateway;
 
-use Illuminate\Support\Facades\Log;
 use Omnipay\Omnipay;
-use Services\PaymentGateway\Dummy;
-use Services\PaymentGateway\Stripe;
-use Services\PaymentGateway\StripeSCA;
 
 /**
  * The intention of this factory is to create a service that is a wrapper around the relative Omnipay implementation
--- a/app/Services/PaymentGateway/StripeSCA.php
+++ b/app/Services/PaymentGateway/StripeSCA.php
@@ -2,8 +2,6 @@
 
 namespace Services\PaymentGateway;
 
-use Illuminate\Support\Facades\Log;
-
 class StripeSCA
 {
 


zcbeaton:
PDF ticket generation is completely broken in this pull request. App\Jobs\GenerateTicket.php no longer works because it is still written for the old PHP facade, which you've replaced. The generatePdfTickets method in App\Models\Order.php also has some problems. For instance, the code which tries to create the PDF tickets directory if it does not already exist instead creates a sub-directory with the name of the order reference. Can you review and fix the ticket generation logic?

CristianDeluxe:
Hello, I have completely separated the two PR so that you can integrate this PR in the master without having to integrate the other, right now you can not apply both at once, I'm waiting to see which of them you apply first to update the other PR.

In principle App\Jobs\GenerateTicket.php has not been modified in this PR, so it should continue to work as it did.

It is true that in the PR of updating the PDF generator I forgot to remove the function generatePdfTickets() from the App\Models\Order.php file, which I just solved, since this function is not called anywhere in the code and can be removed without problems.

If there is any error that causes this specific PR (not counting the other PR of the PDFs) please let me know and I will be happy to fix it.

Thank you for your comments =)

CristianDeluxe:
Tested on two servers and everything is correct at the moment!

publicarray:
Good job on tackling this 👍 I tired checking out your PR, unfortunately the `public/assets/stylesheet/ticket.css` was missing. No problem, I thought, I can generate it with `npm run development` but I got this instead.

```
 ERROR  Failed to compile with 1 errors                                                                                                                                                           9:50:06 pm

This dependency was not found:

* /Users/seb/git/Attendize/resources/sass/ticket.scss in multi ./node_modules/laravel-mix/src/builder/mock-entry.js ./resources/sass/ticket.scss

To install it, you can run: npm install --save /Users/seb/git/Attendize/resources/sass/ticket.scss
 
ERROR in multi ./node_modules/laravel-mix/src/builder/mock-entry.js ./resources/sass/ticket.scss
Module not found: Error: Can't resolve '/Users/seb/git/Attendize/resources/sass/ticket.scss' in '/Users/seb/git/Attendize'
 @ multi ./node_modules/laravel-mix/src/builder/mock-entry.js ./resources/sass/ticket.scss mix[1]
```

Did you miss adding the file to the PR? I also found that the "access codes" page for the organiser returns an exception.

PR #653, opened 2019-06-19
New PDF generator for tickets

CristianDeluxe:
Hello, I have completely modified the way to generate Tickets in PDF, this way is different even from the one I already uploaded in the PR #540 

Now they are generated using Intervention\Image, which allows to have a much nicer and safer design, since PDF generators do not usually parse CSS well and it takes a lot of hours to get an acceptable design in HTML/CSS that is supported by the parser, which can "break" any day.

Generating images ensures that they will always look the same and will not break the design from one version to another, even using other generators in the future.

- Deleted unmaintained package from 2017 nitmedia/wkhtml2pdf
- The pdfs of the tickets are now generated by the new barryvdh/laravel-dompdf, no more permission errors or necessary binaries.
- Deleted unnecessary code, I have avoided duplicating code as much as possible, although there is still work to do on this aspect but will correspond to future PR (such as the code that generates the ticket price... should be a static function accessible from anywhere)

Here is the new ticket design:

Ticket

I suggest to make a new field in the database/form with an exclusive image for the ticket, which would allow graphic designers to adapt better to it, the recommended size would be 1360x635, but this can be done in future PRs, now adapts the flyer image.


Reece0Connor:
I've installed your version to see how well it works.

Is there any reason why you changed to Ticket layout to the displayed version?  (I've attached a pdf ticket that my version spits out)

Overall, I think the new ticket changes are highly welcomed.





[Tickets.pdf](https://github.com/Attendize/Attendize/files/3409338/Tickets.pdf)

CristianDeluxe:
Hi, I think the test you have done has not been with this PR, because this PR generates images, the PDF you uploaded is from my previous PR that was made with HTML converted to PDF and these things happened that you comment (badly formatted). Would you be so kind to try this PR?

scottyeung:
> Hi, I think the test you have done has not been with this PR, because this PR generates images, the PDF you uploaded is from my previous PR that was made with HTML converted to PDF and these things happened that you comment (badly formatted). Would you be so kind to try this PR?

Actually I did, event flyer became shrink up then I needed to comment out the crop part to make it working

RyanAP25:
image
I got this error message while installing this version..

CristianDeluxe:
Please check that you are in the "new-pdf-generator-for-tickets" branch (checkout with git) if you could try it again it would be fantastic

CristianDeluxe:
I close this pull request and open another to integrate it into the next branch.

PR #683, opened 2019-09-09
Added Stripe SCA payment gateway as an option. 

jeremyquinton:
1. Removed dependencies on $gateway->id in the database which made the codebase fairly brittle in places. Use the name field of payment_gatway instead.
2. Refactored Event Checkout so that in theory it should be a lot easier to add other Omnipay gateways.
3. Removed Paypal Express as a gateway supported by default.
4. Payment Gateways Templates are now configurable by specifying an admin and checkout template in the database directly.
5. Broke each payment option out into its own template instead of trying to handle all payment gateways in one form with funny if conditions.
6. Account settings for Payment Gateways is now dynamic based on what is in the database.
7. Added migration and updated Seeder to make sure the payment_gateways is correct on in stall and when upgrading.
8. Updated the refund journeys so that they use the payment Gateway Factory.
9. Improved request data going in and out of the session.

### DIFF
--- a/app/Http/Controllers/EventCheckoutController.php
+++ b/app/Http/Controllers/EventCheckoutController.php
@@ -181,21 +181,16 @@ class EventCheckoutController extends Controller
             ]);
         }
 
-        if (config('attendize.enable_dummy_payment_gateway') == TRUE) {
-            $activeAccountPaymentGateway = new AccountPaymentGateway();
-            $activeAccountPaymentGateway->fill(['payment_gateway_id' => config('attendize.payment_gateway_dummy')]);
-            $paymentGateway = $activeAccountPaymentGateway;
-        } else {
-            $activeAccountPaymentGateway = $event->account->getGateway($event->account->payment_gateway_id);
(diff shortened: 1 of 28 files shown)


CristianDeluxe:
Really good job! Payment processing needed a tune-up!

Under my point of view there is still more work to be done to simplify it further (especially the file EventCheckoutController.php which is crazy) but this is certainly an excellent start.

scottyeung:
The refund journey seems to have problem when dealing with failed payments that shows completed  in Order Status

jeremyquinton:
@scottyeung can you elaborate a bit more on how to replicate the error

scottyeung:
> @scottyeung can you elaborate a bit more on how to replicate the error

Sorry I think I just overlooked...

In regards of Stripe PaymentRequest API, I managed to embed the Apple/Google Pay button into PaymentStripeSCA.blade.php, but this error showed up and I realised a PaymentComplete call is missing, where should I integrate in the frontend.js?

Screen Shot 2019-09-18 at 4 02 18 pm

PR #698, opened 2019-10-15
Fix non-existent partial

CristianDeluxe:
Fixes an error where the blade template tries to load a partial that doesn't exist

### DIFF
--- a/resources/views/Installer/Installer.blade.php
+++ b/resources/views/Installer/Installer.blade.php
@@ -257,8 +257,6 @@
                                     ))  !!}
                     </div>
                     {!! csrf_field() !!}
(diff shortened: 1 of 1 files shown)


jeremyquinton:
@CristianDeluxe 
Seems this partial does exist. 

https://github.com/Attendize/Attendize/blob/master/resources/views/en/Installer/Partials/Footer.blade.php

am I missing something.

etiennemarais:
Thank you 🙏

etiennemarais:
Closing this as the partial does in fact exist. If you look in the localization paths, the `resources/views/<insert_locale_here>/Installer/Partials/Footer.blade.php` exists for all the supported locales. Thank you for your effort though.

PR #700, opened 2019-10-18
Refactored installer controller

CristianDeluxe:
Refactored the installer controller, changed from a "super function" to several functions, removed duplicate code and optimized the order of actions (for example the first thing to check is whether the connection to the database is correct).

I think you could still refactor more, but it's a good start I suppose


CristianDeluxe:
Any news about this?

johannac:
Thank you for contributing this!

Issue #502, opened 2018-10-26
 SQLSTATE[HY000] [2002] php_network_getaddresses: getaddrinfo failed: Name or service not known  


Kibetchirchir:
Iam getting this error when i try 

> sudo php artisan attendize:install

Kibetchirchir:
The problem was on the adress in .env

PR #701, opened 2019-10-18
Update .env.example

hamza-mostafa:
related to issue

https://github.com/Attendize/Attendize/issues/502


etiennemarais:
Thank you 🙏

PR #688, opened 2019-09-20
Upgrade to Laravel 6.5.1

CristianDeluxe:
Here we go again!

I know that there are many files changed but it is necessary to do it well and I think that a "step by step" update would give many errors.

I created a new Laravel project with 6.2 and copied file by file from Attendize, integrating, updating and merging.

It's a big change (a lot of files) but I think it was necessary to start as clean as possible.

You can follow the changes more simply by looking at my commits separately.

- Updated everything to 6.2 and tested everything I could (create organizer, create event, buy ticket, use widget, send email, etc ...)
**Please if someone could do some more tests it would be perfect.**
- Updated all possible composer packages and deleted some unused ones
- Now all routes use the new Laravel 6 syntax.
- Some helpers have been rewritten and now use ServiceProviders
- Deleted some Service Providers and Aliases from config/app.php file because from Laravel 5.5 some have activated the auto discovery function.

Please @jeremyquinton do everything possible to merge this, if I have to change things I will change them, if there is something broken I will fix it :)

I have spent quite a few hours both in this PR and in the two previous attempts, please don't let it be forgotten for months.

My idea with this step is to have a good base to start with the new tests, refactored and simplified classes and of course to fix bugs :)

### Notes:
To avoid errors, after cloning the repository and checking out the "update-laravel" branch, the following steps should be performed:

Install new libraries
`composer install `

Delete old views and cached configurations.
`php artisan config:clear & php artisan cache:clear & php artisan view:clear & php artisan route:clear & php artisan optimize:clear & php artisan clear-compiled & composer dump-autoload`

PR #705, opened 2019-10-23
Makes it possible to process free payments

### LINKED ISSUE #691: Free tickets still requires credit card
I just pulled the latest version and when purchasing a free ticket is requires a credit/debit card.
Any help would be much appreciated.

CristianDeluxe:
Fix #691


jeremyquinton:
@CristianDeluxe thanks for this will look to get this in soon

app/Http/Controllers/EventCheckoutController.php
```diff
@@ -362,7 +362,7 @@ public function showEventPayment(Request $request, $event_id)
             'tickets' => $order_session['tickets'],
             'order_total' => $order_total,
             'orderService' => $orderService,
-            'order_requires_payment'  => (ceil($order_session['order_total']) == 0) ? false : true,
+            'order_requires_payment'  => ceil($order_session['order_total']) > 0,
```
jeremyquinton: I'd prefer to keep this as ternary operator. Reason is in the same file for the same condition a ternary operator is being used. 
https://github.com/Attendize/Attendize/blob/master/app/Http/Controllers/EventCheckoutController.php#L211

Prefer to not add more inconsistencies to an already inconsistent codebase.

CristianDeluxe: Okay, so while we're at it... Why not do it right?

Check my last commit to see if you think it's right:

- Remove (a lot) of duplicate code
- It avoids the cyclomatic complexity of the ternary operator, which is certainly difficult to read.
- Add some helpers to use it in Blade
- Apart from that, it improves the check to support negative numbers, i.e. if the order is -11$ it will say that there is no need to pay it.

We have to do this many times with a thousand more things, but I think it's a good starting point.

etiennemarais:
@CristianDeluxe Thank you for your contribution 🙏 Please can you have a squiz at the conflict?

etiennemarais:
Thanks, I will have a look at this in the morning. Lets get #688 in first and then circle back to this one 👍

PR #726, opened 2019-11-29
Bug fixes

jeremyquinton:
1. Allow Pay offline orders to be cancelled.
2. Fixed bug where if the order requires payment the offline payments option is shown if enabled.
3. Upgrade docker images to PHP7.3
4. Added package.lock.
5. Changed Cancel/Refund button to just cancel to remove ambiguity.

### DIFF
--- a/Dockerfile-composer
+++ b/Dockerfile-composer
@@ -1,4 +1,4 @@
-FROM php:7.1.20-fpm
+FROM php:7.3.12-fpm
 RUN apt-get update && apt-get install -y \
        libpq-dev \
        libmcrypt-dev \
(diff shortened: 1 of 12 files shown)


Dockerfile-composer
```diff
@@ -1,4 +1,4 @@
-FROM php:7.1.20-fpm
+FROM php:7.3.12-fpm
```
etiennemarais: This is great 👍

app/Cancellation/OrderCancellation.php
```diff
@@ -46,8 +46,15 @@ public static function make(Order $order, $attendees): OrderCancellation
      */
     public function cancel(): void
     {
+        $orderAwaitingPayment = false;
+        if ($this->order->order_status_id == config('attendize.order.awaiting_payment')) {
```
etiennemarais:
```suggestion
        if ($this->order->isAwaitingPayment()) {
```

Please extract this part into the `App\Models\Order` model? Then this piece of code can become a bit simpler.

You could drop `$orderAwaitingPayment` as a variable and use the order model method again, which reads a bit easier. 🙏

PR #731, opened 2019-12-16
bugfixes in routes

maartendemandt:
'event_id' was missing from two routes.


etiennemarais:
Thank you for helping with this one 🙏 We appreciate it.

PR #743, opened 2019-12-28
Fix XSS and 500 error if version check fails

publicarray:
The application can execute remote code (XSS) on the Account settings page at the version check. Attendize uses a http request to `https://attendize.com/version.php` to check for a newer version. Since the response is just inserted onto the html any code can be executed. Here is a POC response `9.0.0 <script>alert(1)</script>`

Or just change the lines:

https://github.com/Attendize/Attendize/blob/7ca897ff72c4aa64c42de55d2c351b95d7b42467/app/Http/Controllers/ManageAccountController.php#L231-L242

```php
        $latestVersion = '9.0.0 <script>alert(1)</script>';

        // try {
        //     $http_client = new Client();
        //     $response = $http_client->get('https://attendize.com/version.php');
        //     // Warn: untrusted content! is not validated but inserted into html (EditAccount.blade.phpL149)!
        //     $latestVersion = (string)$response->getBody();
        // } catch (\Exception $exception) {
        //     return false;
        // }
```

I only came across  this issue when I got an 500 error because I blocked outgoing connections. The return value of the function that is getting the current version could return `false`; While there is nothing wrong with that except that the functions calling it never check for this condition. Alternatively the function could have default values set and always return a valid data structure.


publicarray:
Actually visiting the page results in a 404 maybe the version check can be removed instead?

```sh
$ curl -k 'https://attendize.com/version.php'
<html>
<head><title>404 Not Found</title></head>
<body bgcolor="white">
<center><h1>404 Not Found</h1></center>
<hr><center>nginx/1.10.3 (Ubuntu)</center>
</body>
</html>
```

johannac:
@publicarray thanks for this PR! I've gone through the changes and made a few additions. Hope you don't mind I pushed to your branch - I just wanted to get this merged in today so thought it would be easier. 

The original version check file https://attendize.com/version.php is gone now since we redid the website, so I changed it to just do what this old version.php used to do which was get the version file from github.

publicarray:
Thanks yea no problem. And thank you for fixing my buggy code and adding tests 👌

etiennemarais:
@publicarray @johannac Please port these latest changes to master as well? `develop` hasn't been upgraded yet so it will not reflect. We are going to look at archiving that branch soon anyway since we will tag releases and have master as the running main branch.

johannac:
@etiennemarais yes I did rebase the develop branch, so it should be up to date. I was thinking we could use the develop branch as an ‘unstable’ branch and master as a ‘stable’ branch. That way we can merge PRs from contributors to develop and add or change things, if needed, before it goes into master, where we can then tag releases.

PR #744, opened 2019-12-28
Fix XSS and 500 error if version check fails (Next)

publicarray:
Same as #743 for the next branch


etiennemarais:
When this change is merged into `master`, we can merge it into next. You can close this PR since they are the same :)

publicarray:
Thanks, I just wanted to make it easier 👍

PR #774, opened 2020-04-02
Offline Payments Copy Fixes

etiennemarais:
## Context 

If a user selects offline payment when checking out, when you go to the event orders page and click on cancel at the top it says `Cancelling the order will refund the selected attendees`. We shouldn’t show that message for orders which are awaiting payment.

This also counts for cancelling an attendee.

### Changes

- Fixed a couple of named route parameter regressions
- Updated the cancel order to hide refund information if the order used offline payment method
- Updated the cancel attendee to hide refund information if the order used offline payment method

### Screenshots
Screenshot 2020-04-02 at 16 20 33
Screenshot 2020-04-02 at 16 20 42
Screenshot 2020-04-02 at 16 20 51
Screenshot 2020-04-02 at 16 28 02
Screenshot 2020-04-02 at 16 28 07

### DIFF
--- a/routes/web.php
+++ b/routes/web.php
@@ -381,19 +381,19 @@ Route::group(
                 [EventAttendeesController::class, 'postMessageAttendees']
             )->name('postMessageAttendees');
 
-            Route::get('{event_id}/attendees/single_message',
+            Route::get('{attendee_id}/attendees/single_message',
                 [EventAttendeesController::class, 'showMessageAttendee']
             )->name('showMessageAttendee');
 
-            Route::post('{event_id}/attendees/single_message',
+            Route::post('{attendee_id}/attendees/single_message',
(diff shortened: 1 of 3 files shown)


routes/web.php
```diff
@@ -381,19 +381,19 @@
                 [EventAttendeesController::class, 'postMessageAttendees']
             )->name('postMessageAttendees');
 
-            Route::get('{event_id}/attendees/single_message',
+            Route::get('{attendee_id}/attendees/single_message',
```
etiennemarais, line 384: These have regressed from the laravel 6.5 upgrade. The controller methods require an `attendee_id` for each of these routes instead of an event and I think these happened because of copy paste. I tested them again and they seem to work fine

PR #789, opened 2020-04-28
Update ManageAccountController.php

chrisbrosnan:
Corrected 500 internal server error in accessing Account Settings in top right dropdown options, caused by Validator being used when already in use within the same controller (app/Http/Controllers/ManageAccouuntController.php). No problems seem to be caused by this fix in my testing.


chrisbrosnan:
You're very welcome. I'm very glad to!

PR #792, opened 2020-05-04
Add new Functions Clone, Delete,Restore

EstebanJoseSantos:
Routes:
routes.php
Controllers:
EventController: (Adding CreateCloneEvent, archiveEvent, restoreEvent)
Model:
Modify Tickets model,
Views:
ManageEvent, EditTicket, Event, EventPanel


justynpride:
One comment to mention on this, is that when utilized with #803, when you view archived events it displays events from other organizers if the user is Super Admin. Also I get a error 500 page when I clone the event.

bastelix:
can someone review this option?

justynpride:
Yes, please review!

PR #793, opened 2020-05-07
Add hCaptcha

### LINKED ISSUE #740: Combatting Contact form Spam
I'm not an expert on how bots scrape websites, but I know they're finding the contact form on our site and sending a bunch of spam.

I see that the Contact form is just a hidden div. The "Contact" button's click event looks like it just gets rid of the dipaly:none; on the contact form.

I'm thinkin if the form wasn't in the DOM to begin with, bots aren't good at javascript and wouldn't be able to fill in/submit the form. Should I rig up that contact button to generate the form HTML?

I'm sure other folks have had the same issue. Any creative solutions out there?

image

publicarray:
Continuation from #585
Fixes #740

Create a free account on https://www.hcaptcha.com/ and add these 2 settings with your actual keys to your `.env` file.

```
HCAPTCHA_SITE_KEY=aaaa-aaaa-aaaa-aaaa-aaaaaaaa
HCAPTCHA_SECRET_KEY=aaaaaaaa
```


publicarray:
Hi @samdb @johannac, what do you think of using hcaptcha.com ?

johannac:
@publicarray thanks for contributing this and #585! I think, since you've already done some code for both, it would be nice to have support for both and the user can choose via configuration whether to use reCaptcha or hCaptcha or nothing. I will merge this and #585 into develop and then I'll add the parts needed to make them work this way into the develop branch. This will be slated for the July 1st release.

PR #802, opened 2020-05-28
Enhancements to Docker Config

johannac:
This PR adds some enhancements to the current docker config. This is based on the original docker set up for nginx and the docker set up for apache contributed by @Leen15 in #401. 

The goals were:
- to create a single deployable docker image for the web server and worker layers of the application
- support both nginx and apache
- modernize the configuration to take advantage of current docker features (i.e. multi-stage builds)
- streamline the development process a bit

Changes include:
- Added a Makefile including commonly run commands
- Modified 3 nginx Dockerfile's into 1 with multiple build stages
- Both nginx and apache Dockerfile's can be built into a single deployable image, the image can also be used with docker-compose for local development
- Fixed the installer controller which was no longer showing default values in the forms (this is also addressed in a broader scope in #700 which I will merge in later)
- Also added a missing translation to the installer page

Running the local development stack should only require one command now, after cloning the repo, run `make setup` and that's it. Would appreciate feedback from anyone interested in using docker with Attendize, and anyone who would be willing to test the available Makefile commands on your local machine. 

Currently I believe all this only works in Linux based environments, including Mac OS.


jeremyquinton:
This is really good especially for development. A big improvement.

.env.example
```diff
@@ -1,8 +1,8 @@
 APP_NAME=Attendize
-APP_ENV=production
+APP_ENV=local
```
etiennemarais, line 2: 👍

Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+
+# Set up code
+WORKDIR /usr/share/nginx/html
+COPY . .
+
+# run composer, chmod files, setup laravel key
+RUN ./scripts/setup
+
+# The worker container runs the laravel queue in the background
+FROM base as worker
```
etiennemarais, line 14: I love multi stage builds. well done

Makefile
```diff
@@ -0,0 +1,40 @@
+# Commonly used commands, compatible with Linux type shells
```
etiennemarais, line 1: Thank you for adding a makefile. This is a great addition 🚀

PR #811, opened 2020-06-23
Allow ordering free tickets without a payment gateway

SamuelNitsche:
This PR allows ordering free tickets without configuring a payment gateway first.


johannac:
@SamuelNitsche Thanks for submitting this! I can't seem to get it to work on my side. From the event page, when I choose the number of tickets and click the Register button, I get a flash message that there's no payment gateway configured. So I think there might need to be another change in the controller that processes the first step of the checkout flow.

SamuelNitsche:
@johannac You are absolutely right. When testing I enabled offline paying. I updated the checkout controller to only check for a payment gateway if the summed up ticket price is higher than 0.

johannac:
@SamuelNitsche i'm no longer able to merge this due to conflicts (my fault, not yours). So I'm going to close this and I will copy the changes over manually. Thanks!

PR #820, opened 2020-07-07
add translations to locale de

pernpas:


johannac:
Thank you for adding these updates! I assume they are more sensible/contextual translations.

PR #843, opened 2020-10-07
Revert "Add artisan migrate to the setup scripts in case there's data…

trysten:
…base migrations that need to be run"

migrations cannot run on inital setup because sql server has not
started

This reverts commit 8b8c9162f0c226af6b45af66c32939dff64886d5.


trysten:
This PR was premature. The docker build system needs a few more tweaks.

Issue #545, opened 2019-01-03
Unable to edit survey questions (with solution)


scottybo:
Just reviving this very old issue #159 

The solution is pretty simple, in the `EventSurveyController.php` file, in the `postEditEventQuestion` function, modify:

```php
$question->is_required = $request->get('is_required');
```
to be 

```php
$question->is_required = null !== $request->get('is_required') ? $request->get('is_required') : false;
```
So that a value is always set.

philip-loyd:
Did you make a pull request for this? @scottybo

johannac:
This was fixed via https://github.com/Attendize/Attendize/pull/906

Issue #159, opened 2016-07-07
Surveys cannot be edited as non-required


KlausGPaul:
I had set up an event, and a survey. To make the survey applicable to the event, I was selecting "Require this question for ticket(s)..." . 

It was not obvious to me at the time that this would result in me mandatorily requiring all answers to be fille in, I had been assuming this was more of a "also show this survey question on that event". When I tried to edit/disable that setting I got an error response. It does not seem to be possible to change that setting via edit at the moment.

daveearley:
The Survey functionality is very much in its infancy so there's a lot of bugs that need to be ironed out. Are there any messages related to the error in the log file in **Storage/logs**?

Cheers

KlausGPaul:
No, the exported storage/logs is empty, so are the attendize_php and attendize_php-worker container´s storage/logs. Any other area I could check?

Didi I mention that you are doing a great job?

daveearley:
What does the error say? also does the error appear when you are attempting to edit other questions?

Cheers

johannac:
Fixed in #906

PR #862, opened 2020-12-07
Fix for error on save for uncheck survey required

justynpride:
This fixed is referenced in #159 and #545 but it doesn't look like anything was ever submitted to fix it.


justynpride:
Closing as fixed in #906

PR #912, opened 2021-05-07
Event Clone & Archive Feature

justynpride:
This takes the work done in #792 and adds the following fixes so that it full works as expected. 

- when cloning an event an error message occurs when it duplicates tickets. The fix adds the Tickets model to EventsController
- searching on archived events displaying events from other organizers. The fix includes the organizers ID to the search criteria.

I'm still learning Git so couldn't work out how to add it to the previous pull request. It ended up as a pull request on EstebanJoseSantos:master. 

Huge thanks to EstebanJoseSantos for his excellent work in creating the feature.


justynpride:
A helpful addition to this would to be able to delete an event and related tickets etc.. if there are no orders present. This would naturally need a confirmation of some kind before actioning.

PR #916, opened 2021-05-11
Add price to options

justynpride:
This is a work in progress and is based upon the excellent work by @publicarray and pull request #631. This pull request takes that work and makes it work on the latest release 2.2.2. Items that need some final tweaks are: 

1. Display of adding in total on checkout - the correct total that includes the extras is taken, but they aren't shown on the checkout page. All shown correctly on the completed order, emails and dashboard. 
2. Refund for cancelled ticket only does the ticket only, not the extras. 

image
image
image
image

Please feel free to add to this to bring this feature to completion.


michael-heuer:
Hey, I would like to have that feature implemented and would try to help you folks.

Who is in charge and how can I contribute?

Love to hear from you.

Issue #902, opened 2021-04-27
GenerateTicketJob is undefined (wrong use)


quentincaffeino:
https://github.com/Attendize/Attendize/blob/323b60a5304eb6aa0f50d326f4f5be503e3caa29/app/Http/Controllers/EventAttendeesController.php#L6

https://github.com/Attendize/Attendize/blob/323b60a5304eb6aa0f50d326f4f5be503e3caa29/app/Http/Controllers/EventAttendeesController.php#L474

johannac:
This has been fixed in PR #918

PR #918, opened 2021-05-18
Fixing issue #902: wrong import name

nowres:
Fix import name typo


quentincaffeino:
Looks good.

I also see most of the times jobs being dispatched like this:

```php
GenerateTicketJob::dispatchNow($attendee);
```

Instead of how it is now:

https://github.com/Attendize/Attendize/blob/1decf6b4a8fac53bfee25f8fd6c57160712c7ea0/app/Http/Controllers/EventAttendeesController.php#L474

So maybe it would be better to use same style also for this?
But I'm not really a laravel dev so can't tell if there is any real difference.

quentincaffeino (approved):

johannac:
Thank you for submitting this!

PR #937, opened 2021-07-05
added russian translation files

terem42:
This PR adds russian language translation files,


johannac:
This is awesome, thank you for submitting this! We're glad to have Russian language support!

PR #938, opened 2021-07-06
fix for OrderAttendee mail template

terem42:
This PR makes order attendee email sent using correct language translation from lang resources. Currently its hardcoded to English


johannac:
Thank you for submitting this! It looks like I missed that template in the refactoring here  https://github.com/Attendize/Attendize/commit/e6fbc68602408ac3fbbe3436e41d96ec25ee8b62

PR #939, opened 2021-07-06
added locale prefix to email templates calls inside controllers

terem42:
This PR adds much needed prefixes inside controllers to use locale-specific email templates. Without it, all emails are sent in English


johannac:
Thank you for submitting this! Queued emails have language support, but direct send messages did not without this change.

PR #957, opened 2021-08-09
Roles and Permissions: Deactivate, Restore, Delete and send invitations to users

quentincaffeino:
1. Updated to the latest version in develop branch
2. Added ability to Deactivate, Restore and Delete users.
3. Added button to send invitation to user (useful if user forgot his password or his invitation message got lost for some reason)

I called Deactivate as Delete cause I use soft delete functionality. So essentially by deleting user you only mark him as deleted and this makes him unable to use the system. To completely delete user there is a Force Delete button which appears after deleting user.

For user it might be more understandable if Delete will be called Deactivate but I worry that that would cause misunderstanding in code (Deactivate would call delete function). So I'd like to hear some opinions on that matter.

~Also I wasn't sure If I had to push built js to repo.~

Addresses issues mentioned by @justynpride in #803


justynpride:
@quentincaffeino Love this new development. I am probably not doing something, but when I click on the X to deactivate the user, nothing happens, just a red box around the box:

image

Can I ask if you are also testing in a mixed Organiser situation, or just single Organiser? I've found that when testing with more than one Organiser in the system creates errors, especially if you change a users role.

quentincaffeino:
Hey @justynpride, did you rebuilt js, as I've said in the first message I didn't know if built version should go into git or not. For me having built version hosted on git mostly does not make sense. So if it should I can add it no problem.

I think I tested it with multiple organizations but I just accidentally removed myself from super admins so I'll have to fix that and check again)

quentincaffeino:
I did spend some time trying to build js myself and I know that it is not that easy I figured I push prebuild js to make your life easier.

justynpride:
> I did spend some time trying to build js myself and I know that it is not that easy I figured I push prebuild js to make your life easier.

@quentincaffeino Perfect! Huge thanks on this. I have been able to test and it works very well. 

My only issue, and it was there before your work, is concerning the Super Admin user in a multi organization setup. For example:

Organisation 1 has Super User 1 and Organisation 2 has Super User 2. If I login as Super User 2, I get to see the users for Organization 1 as well as being able to switch to the Organization 1 events & profile. Naturally the same is possible if I login as Super User 1, I get to see the users for Organization 2, as well as being able to switch to the Organization 2 events & profile.

A work around is to manually only setup one Super Admin, who then sets up a user to manage events for each organisation. The limitation of this is that each organisation isn't able setup and manage their own users. 

A few options:

1. The Super Admin needs to be limited to just the organizations they are permitted to view, OR
2.  An additional view is given to a User on the **User** role, to view and amend the Users for their organisation, OR
3. A new Organisation Admin role, that is similar to the Super Admin, but that limits access to just the organisations data, and users. 

Next to your new buttons a 'Resend Invite' button would be amazing for those who delete the invite!

quentincaffeino:
I get your concerns but I don't consider them an issue and here is why:
1. it is easier to scale with an 1 "organizer" per instance (organizer is in quotes because I'm talking about a real organizer not a system one)
2. migrating "organizer" is easier since each has a separate database

But I also see how what you are talking about would be useful for many users in the future. I agree with the third option you've brought, where there need to be a global level admin and an organization admin. If @johannac or @etiennemarais do not disagree with those changes I can try make them.

I think I can add `Resend Invite` into this pr to make everyones life easier merging couple of dependent prs or as a new pr when this pr gets merged into `feature/roles-and-permissions`.

justynpride:
Thanks for the helpful feedback. It is definitely a challenging finding something that can work for a number of different use cases. Option 3 was definitely my preferred choice! I hope @johannac and @etiennemarais are happy for these developments and for it to be finally merged in. 

If you are able to do the Resend Invite, I think it would complete the general functions that a normal setup would require. Should the resend only be visible until the user has logged in for the first time? Don't want to over complicate, but also good to simplify the number of buttons on view...

quentincaffeino:
@justynpride added `Send invitation` button. Also put all buttons in dropdown, otherwise they were taking too much space.

@johannac what do you think, can we merge this into `feature/roles-and-permissions`?

quentincaffeino:
@justynpride Although I also don't like how dropdown looks this solution is better because of these two things:

1. Cross character (`&times;`) which was used as a value for both buttons Delete (Deactivate) and Force Delete is very misleading where person could click it and would not realize what it does without looking at title which appears after holding mouse over the button for a second. Which could cause user to accidentally force delete user where they thought they would only disable them.
2. Removing word `Can` from `Can Manage Events` column would not solve the problem that button `Send invitation message` is really long, even if we remove word `message`. And the button is displayed almost always.

But I'm all for more discussion if there are some reasonable arguments for this matter.

justynpride:
@quentincaffeino That all makes sense, thanks for taking the time to explain. It is much appreciated.

justynpride:
@quentincaffeino Just to say that I am finding the additions to this feature so helpful. Thank you! @etiennemarais thanks for your original work on this. I really hope it can be merged in to the main branch soon. @johannac can I ask what the latest thinking is?

johannac:
Hi all - I'm going to be reviewing and testing this branch over the coming weeks. My current plan is to have it fully released by the end of October as our first of the every-two-months feature releases.

PR #970, opened 2021-08-27
Remove googleplus

justynpride:
As Google Plus is no longer an active Google feature, I thought it should be removed from the 'Share' list.


johannac:
Looks great, thanks @justynpride!

justynpride:
@johannac Thanks!

Issue #542, opened 2018-12-18
Stripe Connect!


WASOE:
It will be nice to integrate stripe connect on Attendize, in that way payments will be automatically spilt from ticket vendor to ticket fee 

- https://stripe.com/connect

Mollie has also something like this: https://docs.mollie.com/oauth/overview

vitorpamplona:
This is also important when you have separate bank accounts (Stripe Connect accounts) for each Organizer. It significantly improves the automation in the accounting side.

clawfire:
And since regulation enforce a 2 step authentification on all payments for the 3/4 of the years, the current implementation of stripe will no longer work :/ Stripe had launch a big campaign to collect all the payments solutions editors to reach awareness about this regulation and have everyone ready

samdb:
@clawfire I use Attendize out of the box with Stripe. What are you saying is an issue?

clawfire:
@samdb PSD2 regulation enforces Strong Customer Authentification (SCA) for many online payments. That means the current checkout flow implementer in Attendize doesn't seem to be able to handle dynamic forms of 2FA like 3DSecure v2 (the most common one) or bank customized authentification process (like Revolut, N26, ... uses). 

That mean Stripe will have to comply with the new regulation for making payment and this extra step will be enforced. I'm just suggesting to take the opportunity of the need to adapt the actual payment flow to the new regulation to maybe change the way it's implemented and Stripe Checkout.

More infos : https://stripe.com/en-lu/payments/strong-customer-authentication

clawfire:
hey, now with stripe SCA checkout, we could just add `payment_intent_data` array if we want to use a connected account 😃
Somehting like where `123` is a configurable fee amount you want to keep and the `transfer_data` is the account ID of the connected account. 

```php
'payment_intent_data' => [
    'application_fee_amount' => 123,
    'transfer_data' => [
      'destination' => '{{CONNECTED_ACCOUNT_ID}}',
    ],
  ],
  ```

At first, let people sign-in for connected account can be done manually from the Stripe dashboard (just did it)

justynpride:
@clawfire would the configurable fee be a percentage?  I like the idea and am happy to help test if there is a pull request.

clawfire:
Well the integration allow to do anything. I just did a quick and dirty one on my prod (because I need it for yesterday) and it works well now :p

justynpride:
[...]

PR #972, opened 2021-09-09
Basic support of Stripe Connect 

clawfire:
The idea here is to work on #542 

So I added some elements in the blade StripeSCA config blade to be able to define:

* The transfer account (only support one at that time)
* The fee you want to add (fixed amount so far)

When computing the fee to pass to the transfer account, I sum the Booking fees and the fee defined before. Why? Because when you're doing a connect payment, the main account is supporting the Stripe fee, not the destination account. So I use the Booking fee to cover Stripe's fee and add the custom fixed amount fee.

That's indeed questionable. It works for me, it wouldn't work for everyone. Would be better, probably, to have all of this configured into the Stripe SCA blade under a Stripe Connect Section.

I would be glad to get your feedbacks.


clawfire:
Thanks, @justynpride for the language, I didn't find them 😃 
Added some extra security to test if stripe connect settings are set or not.

PR #935, opened 2021-06-29
upgraded laravel v7

quentincaffeino:
Went through laravel doc and updated laravel to v7.
https://laravel.com/docs/7.x/upgrade

PR #976, opened 2021-09-20
Add support for Mollie

jorisvergeer:
While I was trying to add Mollie support, I found some bugs and other stuff I wanted to change.

Summarized in this PR:
- Add support for Mollie
- Fix bug where request_data and transaction_data is stored as array that gets corrupted when the checkout page is refreshed.
- Change locale in .env
- Don't overwrite .env everytime packages are updated or installed
- Update packages


quentincaffeino:
This is really great work, but I think this should be split into multiple prs because merging big changes is usually just pain. Also it is harder to track changes later.

Some changes as refactors or changing methods (warn -> warning) are not worth doing because this project doesn't have that much tests. Which means that repo owners would need to test these changes manually. This in result might decrease chance of your pr getting merged.

I would split your pr into two: bug fixes and new payment gw. Everything else's I would throw out for better. Or keep as a separate refactor pr if you feel that this is really needed.

justynpride:
@jorisvergeer Should this create a new payment option in the accounts setting? I only get the standard three when working off your commit?

image

Please let me know if I have missed something. I've not done anything with different payment gateways yet, so this is new, and I'm looking to learn. I'd love there to be something for Braintree, which a client of mine uses.

jorisvergeer:
It should create a new one, but I only added it to the existing code that creates the required rows in the database.
I saw that there was one type in one of the seed script. Although I am not sure when and if they are used.

quentincaffeino:
> @jorisvergeer Should this create a new payment option in the accounts setting? I only get the standard three when working off your commit?

@justynpride, have you run installation process after applying this pr? Seeds which create payment gateways are only run on installation (or manually).

justynpride:
@quentincaffeino You are correct. I've rerun the installation and the seeds have come through. Th

justynpride:
New problem now. I've successfully added my test Api key. When testing on the checkout I get this view:

image

rozenlicht:
Is there anything blocking this from getting merged? I'm a new user of Attendize and happy to contribute to the project.

Neustradamus:
Dear @johannac, @egaudrain, @ichier, @jochemp264, @justynpride,

Any progress on it?

justynpride:
There is a helpful discussion on this at https://github.com/Attendize/Attendize/pull/935. 

I've personally moved away from this project, and for any booking system needs use hi.events (https://github.com/HiEventsDev/hi.events). This is in active development and does everything Attendize does and more, and is built on the latest versions of Laravel and Php etc..

PR #1016, opened 2022-02-04
Fix StripeSCA not refunding and charging multiple times

### LINKED ISSUE #940: [BUG] Can't cancel a paid order
**Describe the bug**
When trying to cancel an order (because a client asks for a refund) I cannot process it and got the error message: `You cannot cancel this PaymentIntent because it has a status of succeeded. Only a PaymentIntent with one of the following statuses may be canceled: requires_payment_method, requires_capture, requires_confirmation, requires_action.`

**To Reproduce**
1. Go to orders
2. Select Cancel for an order paid
3. Select the tickets to refund
4. Click cancel. Nothing happens
5. Check the 200 responses of the AJAX Call

**Expected behavior**
Initiate a refund through the payment gateway and cancel the order + attendees

**Version of Attendize you are using**
- tag 2.2.2
- Stripe SCA

**Where you are running Attendize**
- in Gandi Simple Hosting

**What version of PHP you are using on the above server**
- e.g PHP 7.4

**What web server are you using Attendize with**
- Apache.

### LINKED ISSUE #982: Customers charged multiple times
**Describe the bug**
We have now had chance to test this system properly and it is now being used in a live environment where upwards of 10,000 tickets will be sold over the next 6 weeks. We have sold over 800 in last week and have noticed a recurring problem.

We had an order come through and it had repeated itself and charged the customer 4 times. The payment had gone through stripe, new tickets and new orders were created. We thought it was a one off but it has happened at least 6 times now, each time the customers order getting repeated 4 times and being charged 4 times. With the refund / cancel tickets also experiencing a bug where we can not cancel an order all refunds are having to be processed manually and the orders deleted directly from within phpmyadmin.

storca:
Hello!

I'm actively using Attendize for two events, I've ran into this issue where refunding with Stripe SCA payment gateway does not work and hangs on for too long.
We also had issues where Stripe SCA was charging customers multiple times.

### Stripe SCA not refunding

This was the error produced in Stripe 
image

An extract of the error text
```
  "error": {
    "code": "payment_intent_unexpected_state",
    "doc_url": "https://stripe.com/docs/error-codes/payment-intent-unexpected-state",
    "message": "You cannot cancel this PaymentIntent because it has a status of succeeded. Only a PaymentIntent with one of the following statuses may be canceled: requires_payment_method, requires_capture, requires_confirmation, requires_action, processing.",
```

Changing the method from ```cancel``` to ```refund``` solved the problem.

This is the output on stripe after the fix :
image

The refund process in Attendize also works as expected.

Should fix #940 

### Stripe SCA charging customers multiple times

This is my understanding of the problem : when customers click the pay button a request is made to the ```postCreateOrder``` function and during the request is performed, the button is locked.

However, when the request is done loading, the pay button is made active and the client is redirected. 
But redirection can take time, and customers think the payment didn't succeed and click it again.
Another payment request is made during redirection, and this is why customers get charged multiple times.

I've re implemented Stripe elements from scratch and added code to prevent the button from being active upon successful response from the server.

This should fix #982 

There is work to do on the design of the page, but the payment works. 

These two fixes are being actively used in a live environment, and seem to work greatly.

Edit: added the fix for multiple payments with Stripe SCA

Update : works like a charm, 1500+ orders successful, some refunded, multiple charges are gone

### DIFF
--- a/app/Services/PaymentGateway/StripeSCA.php
+++ b/app/Services/PaymentGateway/StripeSCA.php
@@ -104,16 +104,15 @@ class StripeSCA
     public function refundTransaction($order, $refund_amount, $refund_application_fee)
     {
 
-        $request = $this->gateway->cancel([
+        $request = $this->gateway->refund([
             'transactionReference' => $order->transaction_id,
             'amount' => $refund_amount,
-            'refundApplicationFee' => $refund_application_fee,
-            'paymentIntentReference' => $order->payment_intent
+            'refundApplicationFee' => $refund_application_fee
         ]);
 
         $response = $request->send();
 
-        if ($response->isCancelled()) {
+        if ($response->isSuccessful()) {
             $refundResponse['successful'] = true;
         } else {
             $refundResponse['successful'] = false;
--- a/resources/views/Public/ViewEvent/Partials/PaymentStripeSCA.blade.php
+++ b/resources/views/Public/ViewEvent/Partials/PaymentStripeSCA.blade.php
@@ -1,83 +1,148 @@
-<form class="online_payment" action="<?php echo route('postCreateOrder', ['event_id' => $event->id]); ?>" method="post" id="stripe-sca-payment-form">
-    <div class="form-row">
-        <label for="card-element">
-            @lang("Public_ViewEvent.stripe_credit_or_debit_card")
-        </label>
-        <div id="card-element">
-
-        </div>
-
-        <div id="card-errors" role="alert"></div>
-    </div>
-    {!! Form::token() !!}
-
-    <input class="btn btn-lg btn-success card-submit" style="width:100%;" type="submit" value="@lang("Public_ViewEvent.complete_payment")">
-
+<form class="online_payment" id='payment-form'>
+    <label>
+      Card details
+    </label>
+    <!-- placeholder for Elements -->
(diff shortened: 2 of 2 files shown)


storca:
Maybe my fix on Stripe SCA multiple payments is a bit aggressive, I think it's possible to stick a little bit more to the old code base, I'll probably do this when I have the time to do so

clawfire:
@storca maybe splitting in different PR will increase the chances to get validated. Like ... it could make your fix of #940 pass while you're working on #982

storca:
I mean, most of my changes are minor...
If there's any disagreement with my changes we can work it out, plus creating two PRs is a nightmare given the current state of the commits.

johannac:
Thank you @storca for submitting this fix! I'm going to merge this in and then clean up a few things if that's ok. The majority of it looks great, but a few things could be cleaned up with the js and css for the long term. I'll also incorporate the patch from @ghoussay, I noticed the lang calls got dropped as well. 

Thanks again all! New release coming shortly.

storca:
Sorry for the delay

@ghoussay the fixes have been tested for two real events, mentionned in #1027 

@johannac absolutely that's okay! It was a quick fix for me so there was things to fix :)

PR #1051, opened 2022-09-18
Capitalisation error

justynpride:
Correct so Order attendees now displays as Order Attendees in keeping with the other style on the page.


johannac:
Thank you @justynpride!

Issue #1036, opened 2022-07-13
We'll be back soon - Logo Path Error 


AskSabry:
Hello, 
I have a We'll be back soon page when i use logo for Event Page 
and when put in debug mode 
it says that 
is not within the allowed path(s):
resources/views/Public/ViewOrganiser/Partials/OrganiserHeaderSection.blade.php)

    public function getFullLogoPathAttribute()

    {

        if ($this->logo_path && (file_exists(config('attendize.cdn_url_user_assets') . '/' . $this->logo_path) || file_exists(public_path($this->logo_path)))) {

            return config('attendize.cdn_url_user_assets') . '/' . $this->logo_path;

        }

        return config('attendize.fallback_organiser_logo_url');

I'm in Apache 

thanks

justynpride:
I've not got a solution for this, but I encounter the same issue. The error in my log is:

`[2022-07-13 21:44:52] production.ERROR: file_exists(): open_basedir restriction in effect. File(/user_content/organiser_images/book-inme-logo-1.png) is not within the allowed path(s): (/home/bookinme/.composer:/home/bookinme/web/tickets.book-in.me/public_html:/home/bookinme/web/tickets.book-in.me/private:/home/bookinme/web/tickets.book-in.me/public_shtml:/home/bookinme/tmp:/tmp:/var/www/html:/bin:/usr/bin:/usr/local/bin:/usr/share:/opt) (View: /home/bookinme/web/tickets.book-in.me/public_html/resources/views/Public/ViewOrganiser/Partials/OrganiserHeaderSection.blade.php) {"userId":1,"exception":"[object] (Facade\\Ignition\\Exceptions\\ViewException(code: 0): file_exists(): open_basedir restriction in effect. File(/user_content/organiser_images/book-inme-logo-1.png) is not within the allowed path(s): (/home/bookinme/.composer:/home/bookinme/web/tickets.book-in.me/public_html:/home/bookinme/web/tickets.book-in.me/private:/home/bookinme/web/tickets.book-in.me/public_shtml:/home/bookinme/tmp:/tmp:/var/www/html:/bin:/usr/bin:/usr/local/bin:/usr/share:/opt) (View: /home/bookinme/web/tickets.book-in.me/public_html/resources/views/Public/ViewOrganiser/Partials/OrganiserHeaderSection.blade.php) at /home/bookinme/web/tickets.book-in.me/public_html/app/Models/Organiser.php:118)
`

qlickit:
[...]

PR #1061, opened 2022-10-24
fix open_basedir restriction

ichier:
open_basedir in effect currently breaks organiser-page and ticket creation
i just swapped file-checks to test for the local one first
see https://github.com/Attendize/Attendize/issues/1036#issuecomment-1288177215


johannac:
@ichier Thank you for contributing this!

PR #1063, opened 2022-10-26
Add organiser logo to confirmation emails

egaudrain:
Receiving emails with the Attendize logo can be confusing for attendees.

The email Layout file [resources/views/Emails/Layouts/Master.blade.php](https://github.com/Attendize/Attendize/blob/develop/resources/views/Emails/Layouts/Master.blade.php#L25) makes reference to `$email_logo` but falls back to `assets/images/logo-dark.png` if it is not defined.

This pull request adds the `email_logo` property to the objects passed to the view engine.

Note 1: The master layout uses `isset($email_logo)`. I am not sure what happens if the organiser did not define any logo.
Note 2: I tested SendOrderAttendeeTicketMail.php and SendOrderConfirmationMail.php, but not the three others.


johannac:
Thanks for contributing this @egaudrain! I will merge and test the other email templates.

Issue #1027, opened 2022-05-27
[FEEDBACK] Used Attendize for two big events!


storca:
I've set up Attendize for two big events, one at ~1000 people, the other one was at ~6000 tickets sold.
In those two events, I needed to use the branch `feature-roles-and-permissions` for the volunteers to scan tickets without accessing other functionalities of Attendize.

Here's what came back of it, ordered from the most problematic, to the least

* **Disable development mode in production...**
Don't forget it please...

* The scan of the QR Codes highly depends on the conditions of the day !
If there's sun, if the phone's screen is broken **it doesn't work**!
Scan on printed tickets works.
Ideas : Change the QR Code scanning library ; develop some kind of hardware for the 1D barcodes, with the API *It would be fun to do*

* The scan page lags
The scan page is a bit unstable : out of the +50 people that scanned tickets, it was clear that the page worked for some time, but then is starts lagging quite a lot (delays when pressing the QRCode button, the image of the camera lags, stops, stays black, ...)
There may be maintenance work to do around here!

* Access codes
It would have saved me **hours** (not joking) if there was a limit on how much tickets can a access code order.

* PDF documents
Questions were very usefull! But uploading documents would be a great addition, we lacked this functionality a lot and had to work with Google Drive.
Idea : As in Google's own file system, a unique time-based hash is generated for each file, file is renamed to this hash and is publicly accessible. The file's hash is set as `answer_text` in `question_answers`

**Attendize is great sotware!** *It just needs a bit more attention*
It's a bit buggy sometimes **BUT** these minor issues can be worked around easily and don't compromise the overall function of Attendize.
Attendize truly saved me a lot of time with the refunds, once my patch with Stripe was applied.

I'll add more things here if I remember more

Thanks again to all the community of contributors who helped in any way!

rootjack404:
Couldn't agree more in the point number 2, it's really hard to scan in outdoor event, sometime because the reflection, sometime because the customer screen is too dark, and asking attendee to print the ticket is not the option in the current day it's just feel too unnecessary in the customer side, and make waste problem even worse.

justynpride:
@storca tha is for the helpful feedback. 8its encouraging to hear that it served your events and scale of numbers well. Out of interest where there many ‘survey questions’  being asked? For me we have a number of questions being asked, and naturally need go then amend so answers down the road. This is a challenge for me as the system currently operates,

storca:
@justynpride if I understand correctly, you have trouble with processing the answers

[...]

PR #1067, opened 2022-11-12
Change QrCode scanning to nimiq/qr-scanner and improve scan UI readability

storca:
Hello!

Following issue #1027 regarding the issues I've had with Qr Code scanning, I'd like to change the legacy QrCode scanning library to [nimiq/qr-scanner](https://github.com/nimiq/qr-scanner).

What I've done :
- Changed the scanning library to qr-scanner
- Adapted blade template, CSS and JS to fit the library
- Enhanced UI readability when scanning tickets (clearer distinction between what's important and what's not)

Note : I've had to change the way the video is displayed in the check-in page, the video preview is now smaller.

Let me know what you think!


storca:
It appears that the video is not responsive and does not display correctly on smaller screen sizes.
I'm working on the issue but my skills are limited in CSS, help would be appreciated.

storca:
Here are the scan UI previews after the video preview responsiveness fix :

On a Pixel 6 using Chrome :
video preview
scan result

On a wide screen :
video preview
Scan result

storca:
PR bump

This PR also fixes a small inconsistency in the scan UI : the order reference was always uppered when displayed, but the order reference is case-sensitive. This is fixed.

I also changed the font to a Mono font to better distinct characters l, I and 1.

rootjack404:
Hi,
Thanks for your great work!
anyway, I want to ask how do I pass the inversionMode on this? Since I had black background design ticket

PR #1074, opened 2022-12-23
Add file upload question, validation, storage, and preview

### LINKED ISSUE #1027: [FEEDBACK] Used Attendize for two big events!

### LINKED ISSUE #1041: [FEATURE] Atendee must upload a file before getting a ticket
Hi,

First thank you for the work this tool looks great

I would like to use this tool for a school, i have installed the testing version using docker.
Everything seems to be working so far.
I only have one thing that i am not able to do and i am not sure if this tool lacks the feature or if it is me that don't know how to use it.

I have two type of event : free and not free
the ones that requires a payement are not an issue with this tool
the free one in the other hand are giving me headaches.
What i need is that atendees have to upload a file while getting a ticket
I also need to be able to export all the uploaded files and to know wich files has been emitted by who and from wich event

i dont know if it is possible to do that yet ? 

what i think is to use survey, i saw that we can make thoses a requierement for atendee to book a ticket
I might code it myself so any help on where the code for survey is would be apreciated !

if this feature is available on the non dev version please enlighten me :D

Thanks in advance for your help

storca:
Hi!

Following issues, I added a file upload question in Attendize surveys (fixes #1041 and fixes #1027).

How does it work ?

- When users submit files, they are validated by : mimetype, size and the 'required' flag of the question
- If files are validated, they are stored in a temporary directory
- If the order is successfull, files are moved in a permanent location
- The partial path to the uploaded documents is stored in ```answer_text``` in the ```question_answers``` table like so ```docs/<custom_hash>/<laravel_file_hash>.ext```

> Publicly accessible ??

Yes, but the files are hidden via the following path structure ```/docs/<custom_hash>/<laravel_file_hash>.ext```

The ```custom_hash``` is made in an effort to make sandwich-proof hashes in case it is possible to get hashes during the order process.

The hash is generated using php time and OS-based randomness [here](https://github.com/storca/Attendize/commit/a8aab86c340afff321e607221f8e8ac366fcc2b4#diff-68a094e6975d52e95452423f68eab7168d38e01469fabbd87b4b2d0a0cd2eceaR726).

Files are not accessible or editable by the user.

Files can also stay on the server in case the order is cancelled or abandoned ; this is why I added a ```RemoveOldTempFiles``` job that removes files that are older than an hour.

I'm happy to share this PR with you, feel free to test it and suggest changes!

### DIFF
--- a/app/Http/Controllers/EventCheckoutController.php
+++ b/app/Http/Controllers/EventCheckoutController.php
@@ -362,24 +362,26 @@ class EventCheckoutController extends Controller
                 'messages' => $order->errors(),
             ]);
         }
-
-        // All files are validated, now remove file objects from array to avoid serialization errors
-        // Files are stored in a temp folder, awaiting order confirmation
-        $order_hash_data = microtime().random_bytes(10);
-        // take only the first 20 characters, it would take 16^20 bruteforce requests
-        $folder_name = substr(hash('sha256', $order_hash_data), 0, 20);
-        $order_dir_path = 'docs/'.$folder_name;
-        //File::makeDirectory($order_dir_path);
-        foreach($request_data['ticket_holder_files'] as $key => $ticket)
+        if(isset($request_data['ticket_holder_files']))
         {
-            foreach($ticket as $i=>$item)
+            // All files are validated, now remove file objects from array to avoid serialization errors
+            // Files are stored in a temp folder, awaiting order confirmation
+            $order_hash_data = microtime().random_bytes(10);
+            // take only the first 20 characters, it would take 16^20 bruteforce requests
+            $folder_name = substr(hash('sha256', $order_hash_data), 0, 20);
+            $order_dir_path = 'docs/'.$folder_name;
+            //File::makeDirectory($order_dir_path);
+            foreach($request_data['ticket_holder_files'] as $key => $ticket)
             {
-                foreach($item as $question_id => $file)
+                foreach($ticket as $i=>$item)
                 {
-                    if(isset($file)) {
-                        // Store the file in a temporary folder, use Laravel default file name
-                        $path = $file->store("tmp/$order_dir_path");
-                        $request_data['ticket_holder_files'][$key][$i][$question_id] = $path;
+                    foreach($item as $question_id => $file)
+                    {
+                        if(isset($file)) {
+                            // Store the file in a temporary folder, use Laravel default file name
+                            $path = $file->store("tmp/$order_dir_path");
+                            $request_data['ticket_holder_files'][$key][$i][$question_id] = $path;
+                        }
                     }
                 }
             }
--- a/app/Http/Controllers/EventCheckoutController.php
+++ b/app/Http/Controllers/EventCheckoutController.php
@@ -700,6 +700,7 @@ class EventCheckoutController extends Controller
                 /*
                  * Create the attendees
                  */
+                $tmp_dir = null;
                 for ($i = 0; $i < $attendee_details['qty']; $i++) {
 
                     $attendee = new Attendee();
@@ -726,7 +727,7 @@ class EventCheckoutController extends Controller
                         $file_hash_data = microtime().random_bytes(10);
                         $foldername = substr(hash('sha256', $file_hash_data), 0, 20);
                     }
-
+                    
                     foreach ($attendee_details['ticket']->questions as $question) {
                         // If there is a temp file there
                         if(isset($request_data['ticket_holder_files'][$attendee_details['ticket']->id][$i][$question->id])) {
@@ -735,10 +736,10 @@ class EventCheckoutController extends Controller
                             $definitive_path = "docs/$foldername/".basename($tmp_path);
                             // Move the file
                             Storage::disk('local')->move($tmp_path, $definitive_path);
-                            // Remove the temp directory
-                            Storage::disk('local')->deleteDirectory(dirname($tmp_path));
                             // Change file path
                             $ticket_questions[$attendee_details['ticket']->id][$i][$question->id] = $definitive_path;
+                            // Set tmp_dir
+                            $tmp_dir = dirname($tmp_path);
                         }
 
                         $ticket_answer = isset($ticket_questions[$attendee_details['ticket']->id][$i][$question->id])
@@ -769,6 +770,10 @@ class EventCheckoutController extends Controller
                     /* Keep track of total number of attendees */
                     $attendee_increment++;
                 }
+            }
+            if(isset($tmp_dir)) {
+                // Remove the temp directory if needed
+                Storage::disk('local')->deleteDirectory($tmp_dir);
             }
         } catch (Exception $e) {
             Log::error($e);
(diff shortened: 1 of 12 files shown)


justynpride:
Love this! Thank you for your excellent work. It works just as I would expect it to. One thought for a feature improvement would be for when you view the survey answers. If you could click on the file name and view it in a new window or modal would be amazing.

justynpride:
@storca Just checking -  have you run this on a fresh install? I ask as I am getting a sql error when doing a fresh install with this. happy for it to be an error at my end....

SQLSTATE[23000]: Integrity constraint violation: 1062 Duplicate entry '1' for key 'PRIMARY' (SQL: insert into `question_types` (`alias`, `allow_multiple`, `has_options`, `id`, `name`) values (text, 0, 0, 1, Single-line text box), (textarea, 0, 0, 2, Multi-line text box), (dropdown, 0, 1, 3, Dropdown (single selection)), (dropdown_multiple, 1, 1, 4, Dropdown (multiple selection)), (checkbox, 1, 1, 5, Checkbox), (radio, 0, 1, 6, Radio input))

justynpride:
> 

I think it just needs 'id' => 7, added to the migration file.

storca:
@justynpride Yes the new question id is 7 on a migrated database, if yours differ, you can always change the question id in config/attendize.php line 62.

justynpride:
Thanks. Without that I'd line I think it defaults to 1 and therefore created a duplicate install error in sql.

The view file link is really helpful by the way.

justynpride:
@storca I realised in testing I only tested orders with the file question on the ticket. I'm finding that if I don't have question on a ticket that when I try to move to the Payment section I get a Whoops Error. The log shows:

[2023-01-01 09:44:15] production.ERROR: Undefined index: ticket_holder_files {"userId":1,"exception":"[object] (ErrorException(code: 0): Undefined index: ticket_holder_files at /home/bookinme/web/bookings.book-in.me/public_html/app/Http/Controllers/EventCheckoutController.php:373)

I don't know if you are able to replicate on a clean install of 2.6.0?

justynpride:
@storca Perfect, it works! Related to this with your helpful 'Open file' link on the survey question, do you get to view the uploaded file ok? I've been trying it more (apologies I had just view how it looked previously!), and just this on a Mac:

Screenshot 2023-01-05 at 20 25 18

johannac:
@storca how do you see this feature being used? So like if an event needs a photo ID or copy of your covid vaccine card, you could upload a photo of it? That kind of thing?

storca:
@justynpride the preview of the documents relies on browser-based document preview, does it work for you ? What type of documents do you upload ?

@johannac yep, in our case we use Attendize for a sports tournament and we need to ask our participants for documents like medical certificates

justynpride:
> 

@storca The preview works fine for a pdf, but not a jpg or png. It would be likely that images would be uploaded, and not just  pdfs.

storca:
@justynpride I'm very surprised that your browser does not allow the preview of images, I can't provide troubleshooting but try to have a look at the code below and fix it yourself.

https://github.com/storca/Attendize/blob/a5cdaf1e2c81b61c310df6357944e4858ae92898/app/Http/Controllers/EventCheckoutController.php#L899-L910

justynpride:
@storca It is weird as I run MacOS and Safari and can't view the image previews on either MacOS or iOS. before looking more widely can I just check what devices you know it works with?

storca:
@justynpride I use Chromium or Firefox on Linux, try changing your browser

justynpride:
> @justynpride I use Chromium or Firefox on Linux, try changing your browser

I also tend to try out errors on other browsers. The same occurs in Chrome and Firefox for images (fine for pdfs). In Firefox I get this error displayed:

The image "https://tickets.book-in.me/docs/d7dd181aba01d76fbba6/z6zTVyU4vDYNz1DYVOp4ok5FR0wXLZ3S0h270Ran.png" cannot be displayed because it contains errors.

storca:
@justynpride yes indeed, it contains errors ; are you sure that your image is not corrupted ?
I'm using this feature with PDFs, PNGs and JPEGs and it works fine ; try opening a PNG image with your browser, if it displays it then upload it to attendize and try previewing it. It should work.

justynpride:
@storca Unfortunately the image opens fine in the browser (ie I drop the file onto the app and it opens within the browser window), but not when previewing through Attendize. I need to try on a PC when I have access.

app/Http/Controllers/EventCheckoutController.php
```diff
@@ -819,4 +896,16 @@ public function showOrderTickets(Request $request, $order_reference)
...
 
+    public function showOrderDoc($folder_name, $file_name)
+    {
+        $fp = Storage::disk('local')->path('docs').'/'.$folder_name.'/'.$file_name;
+        if(is_file($fp)) {
+            return response()->file($fp);
+        }
+        else {
+            // Prevent bruteforce
+            sleep(1);
```
publicarray, line 907:
```suggestion
```

https://www.php.net/manual/en/function.sleep.php

> Not really - the attacker could do 100 requests. Each request might take 2 seconds but it doesn't stop the number of requests done. You need to stop processing more than one request every 2 seconds rather than delay it by 2 seconds on each execution.

This can be done on the webserver e.g. nginx or using laravel's own Reatelimiting on the route:
https://laravel.com/docs/10.x/routing#rate-limiting

With sleep you also reduce the amount of requests the CPU could be serving at this time.

